PR #13, opened 2022-09-03
Feature/health check

### LINKED ISSUE #6: global: create endpoint to check the server status
Currently, the root path of the server (`/`) provides and `index.html`, which doesn't make sense for a REST API. The goal of the issue is to delete this static file and instead implement a simple endpoint that just returns and 200 status code and a simple json object informing the server is online and running

bdmendes:
Closes #6 
Adds an health check at the root endpoint, and an appropriate error message for undefined routes. Adds a base `README` file for later expansion.


src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+class ErrorController : ErrorController {
+    val errorKey : String = "error"
+
+    @RequestMapping("/**")
+    fun endpointNotFound(): Map<String, String> = mapOf(errorKey to "invalid endpoint")
+}
```
BrunoRosendo: No empty line (I guess the linter would say something about this)

src/main/kotlin/pt/up/fe/ni/website/backend/controller/IndexController.kt
```diff
@@ -0,0 +1,10 @@
+package pt.up.fe.ni.website.backend.controller
+
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+class IndexController {
+    @GetMapping("/")
+    fun healthCheck(): Map<String, String> = mapOf("online" to "true")
+}
```
BrunoRosendo: No empty line

src/main/resources/application.properties
```diff
@@ -1,9 +1,13 @@
...
+
 # Datasource
 spring.datasource.url=jdbc:h2:file:~/website-be-h2-db
 spring.datasource.username=h2dev
 spring.datasource.password=
 spring.datasource.driver-class-name=org.h2.Driver
 spring.jpa.hibernate.ddl-auto=update
+
+# API Settings
+server.error.whitelabel.enabled=false
```
BrunoRosendo, line 13: What is this for? I couldn't understand the difference when removing this

BrunoRosendo (changes requested):
Good job. Some minor comments

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
bdmendes: Indeed the linter (now merged from develop) warns about this. Fixed

src/main/resources/application.properties
bdmendes: This makes the server not present the default HTML error page for undefined endpoints, instead falling back to the Apache Tomcat 404 page. The latter behavior is disabled, though, by the `ErrorController`, which sends a json containing "Invalid endpoint". (so, in practice, it has no effect for us, but makes sense to me since we're building an API)

BrunoRosendo (approved):

PR #16, opened 2022-09-04
Reorganize DB configuration

### LINKED ISSUE #2: db: investigate which DB and ORM to use
Right now, we're using JPA and a simple H2 database for development. We should decide if we want a SQL or Mongo DB and whether it makes sense to keep using JPA or not

BrunoRosendo:
Closes #2 

After some discussion regarding which DB to use, we decided to stick with SQL. For development, we can keep using H2 and, for production, we only need to change the `spring.datasource.url` configuration in `application.properties`. This PR also removes the explicit driver from the config, since Spring Boot already infers it correctly based on the URL

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/EventController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/EventController.kt
@@ -12,7 +12,7 @@ import pt.up.fe.ni.website.backend.service.EventService
 @RequestMapping("/events")
 class EventController(private val service: EventService) {
     @GetMapping
-    fun getAllEvents(): Collection<Event> = service.getAllEvents()
+    fun getAllEvents() = service.getAllEvents()
 
     @PostMapping("/new")
     fun createEvent(@RequestBody event: Event) = service.createEvent(event)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/pt/up/fe/ni/website/backend/controller/EventController.kt
```diff
@@ -12,7 +12,7 @@ import pt.up.fe.ni.website.backend.service.EventService
 @RequestMapping("/events")
 class EventController(private val service: EventService) {
     @GetMapping
-    fun getAllEvents(): Collection<Event> = service.getAllEvents()
+    fun getAllEvents() = service.getAllEvents()
```
BrunoRosendo, line 15: I think this makes the functions more readable, since they're very simple. Do you agree?

bdmendes: I think it is fine, maybe some will disagree tho

BrunoRosendo: Can we close this then?

limwa: There is a slight difference between having the type annotation and not having one. Without a type annotation, the return type of the function will be List<Event!> instead List<Event>. The exclamation mark means that the IDE doesn't know if the type is nullable. We should avoid these things to ensure as much null safety as possible, thus, writing out the type, at least in this case, is a good practice, in my opinion

bdmendes (approved):

Issue #15, opened 2022-09-04
errors: add global error handling


BrunoRosendo:
We should be able to handle general errors globally. For example:
- Invalid endpoints (already implemented)
- Invalid elements accessed by endpoints like `/events/{eventId}`. We should be able to handle this globally with an `@ExceptionHandler` on the `NuSuchElementException`, you have an example in the final minutes of [this video](https://www.youtube.com/watch?v=jiv37i503v4)
- Invalid body when creating resources. For example, when creating an event, if the request is missing one field (e.g. title), you receive an `invalid endpoint` error, which is confusing. We should try to return information about the missing field.
- Any other errors you may see fit

It would also be nice to normalize the errors and have a consistent format, making it easier to handle them in the FE. For example, we could use a JSON format like {"message": "error"} or something similar to this.
You can probably extend the `ErrorController` that already exists

BrunoRosendo:
Check cases when a required field is set as null in the request body. This is currently being ignored

PR #19, opened 2022-09-09
10 posts create model and endpoints

### LINKED ISSUE #10: posts: create model and endpoints
Posts should provide quick information about the branch's activity and link to a project or event.
A rich text solution can be investigated.

AvilaAndre:
Closes #10

The Post object has a string _description_, a string _link_, a date _date_ and a long _id_.
This PR implements a **GetMapping** for all posts at "/" for a single post at "/{postID}" according to the _postID_, it implements a **PostMapping** creating a new Post object at "/new" receiving in the request body the _description_, _link_ and _date_, it implements a **PatchMapping** at "/{postID}" updating the _link_, _description_ or both (specified on the request body) for the Post with the given _postID_, and lastly, it implements a **DeleteMapping** at "/{postID}" deleting the Post with the given _postID_.


BrunoRosendo:
@AvilaAndre I haven't reviewed yet but you might want to already fix the linting. You can look at the errors by running `./gradlew ktlintCheck` or checking the GitHub action. We can maybe explain in the next meeting how to setup the linter on your IDE, to ease the process

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,33 @@
+package pt.up.fe.ni.website.backend.service
+
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.stereotype.Service
+import pt.up.fe.ni.website.backend.model.Post
+import pt.up.fe.ni.website.backend.repository.PostRepository
+
+@Service
+class PostService(private val repository: PostRepository) {
+    fun getAllPosts(): Collection<Post> = repository.findAll().toList()
```
BrunoRosendo:
```suggestion
    fun getAllPosts(): List<Post> = repository.findAll().toList()
```

Me and @limwa were talking and we think it's better to pass `List` here, since it has more methods and could be more useful in the controllers

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    fun getPost(postID : String): Post = repository.findByIdOrNull(postID.toLong()) ?: throw NoSuchElementException("Post Not Found")
```
BrunoRosendo: I think postID should be a Long here and in the controller. Is there any reason I'm not seeing to have it a String?

BrunoRosendo: If you decide to use Long here, please change in other places as well

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,33 @@
...
+        if (newDescription != null) {
+            targetPost.description = newDescription
+        };
+        if (newLink != null) {
+            targetPost.link = newLink
+        }
+
+        repository.save(targetPost)
+
+    }
```
BrunoRosendo: This kind of partial updates are a lot more tricky than I initially thought, I feel like Spring Boot doesn't have a perfect way to do it. This is definitely the simplest way to do it but I'm afraid it'll be hard to scale and will loose readability. I will leave some alternatives here but I'm not sure of the best one myself, I'll try to give it some more thought:

- Maintain this logic but make use of Kotlin's operators:

Instead of writing `if (newLink != null) {targetPost.link = newLink}` you can do `targetPost.link = newLink ?: targetPost.link` or even `newLink?.let { targetPost.link = newLink }`. See what you prefer

- Reduce the number of parameters

Instead of passing all the fields as parameters, we could maybe simply pass a `Post` and check if the fields are not null. Including the changes in last point, it'd be something like this (I hope this doesn't bring problems with the partial requests):
```suggestion
    fun updatePost(postID: String, post: Post) {
        val targetPost = repository.findByIdOrNull(postID.toLong()) ?: throw NoSuchElementException("Post Not Found")
        targetPost.description = post.description ?: targetPost.description
        targetPost.link = post.link ?: targetPost.link

        repository.save(targetPost)

    }
```

- Use a Data Transfer Object

Some projects use DTO's as the practice for partial updates. They are basically a model with only nullable fields that can be updated, and they're then mapped to a given object. At first sight, I don't quite like them because we'll get another class for each model we need to update, but maybe it's worth for you to check it out. You can read about it here: https://www.baeldung.com/spring-data-partial-update

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.*
+import javax.persistence.Entity
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+
+@Entity
+class Post (
+    var description: String,
+    var link: String,
```
BrunoRosendo: I'm wondering what this `link` is supposed to be used for?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
```diff
@@ -0,0 +1,32 @@
...
+import pt.up.fe.ni.website.backend.service.PostService
+import kotlin.NoSuchElementException
+
+@RestController
+@RequestMapping("/posts")
+class PostController(private val service: PostService) {
+
+    @ExceptionHandler(NoSuchElementException::class)
+    fun handleNotFound(e : NoSuchElementException): ResponseEntity<String> =
+            ResponseEntity(e.message, HttpStatus.NOT_FOUND)
```
BrunoRosendo: We should handle these exceptions globally later on. We can leave it for #15

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    fun deletePost(postID : String) {
+        val targetPost = repository.findByIdOrNull(postID.toLong()) ?: throw NoSuchElementException("Post Not Found")
+        repository.delete(targetPost)
+    }
+}
```
BrunoRosendo: This works nicely and I'm just leaving this comment for the record. We should probably find a nice way of validating the `postID` whenever it's used and throw `NoSuchElementException("Post Not Found")` there. That way, we could simplify this to something like this. We could think about it in #15 too
```suggestion
    fun deletePost(@ExistingPost postID : Long) {
        repository.deleteById(postID)
    }
```

BrunoRosendo (changes requested):
Nice job so far!
I wrote many comments but most of them are cosmetical and I opened some discussions for future improvements

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
AvilaAndre: I don't have a clear idea of what the Post should be, therefore I just thought it would be like the social media posts but with the link to the resource mentioned, for example, a post about TTS could have the link _ni.fe.up.pt/tts/planner_ or link to the page on the website dedicated to the project.

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
AvilaAndre: I thought it had to be a string, already refactored it.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
BrunoRosendo: I get what you're thinking. As for the model fields, this is my opinion:
- I understand the posts are usually about something (like a project or an event) but I don't think that'll always be the case. I think the links should simply be made with an `href` (assuming the post's body will be HTML, it can be something else but the concept still applies)
- I think `description` should be changed to `body` and we should add a `title` field. That makes more sense to me for a post.
- While we're at it, we could change `date` to `publishDate`, just to be more clear
- I would consider adding a template (a string called `templatePath` or something like that)

What do you think?

src/test/kotlin/pt/up/fe/ni/website/backend/BackendApplicationTests.kt
```diff
@@ -6,5 +6,6 @@ import org.springframework.boot.test.context.SpringBootTest
 @SpringBootTest
 class BackendApplicationTests {
     @Test
-    fun contextLoads() {}
+    fun contextLoads() {
+    }
```
BrunoRosendo, line 10: Was this a change from the linter? Looks weirder to me but if it is then just leave it

src/main/kotlin/pt/up/fe/ni/website/backend/model/PostDTO.kt
```diff
@@ -0,0 +1,9 @@
+package pt.up.fe.ni.website.backend.model
+
+import java.util.Date
+
+class PostDTO {
+    val description: String? = null
+    val link: String? = null
+    val date: Date? = null
+}
```
BrunoRosendo: In order to avoid bloating the model with DTO's, I think this class should just be in the `PostService.kt` file since it's intrinsic to the service. I know we usually follow a "one class per file" approach in Java but I don't think we need to follow it here in Kotlin and it seems more organized to me

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,31 @@
...
+    fun getPost(postID: Long): Post =
+        repository.findByIdOrNull(postID.toLong()) ?: throw NoSuchElementException("Post Not Found")
+
+    fun createPost(post: Post) = repository.save(post)
+
+    fun updatePost(postID: Long, post: PostDTO) {
+        val targetPost = repository.findByIdOrNull(postID) ?: throw NoSuchElementException("Post Not Found")
+        targetPost.description = post.description ?: targetPost.description
+        targetPost.link = post.link ?: targetPost.link
+        targetPost.date = post.date ?: targetPost.date
```
BrunoRosendo: Why is the date editable?

BrunoRosendo (changes requested):
This is looking good. I like this way of using the DTO more than using a mapper.

I left some comments regarding posts' fields and cosmetics

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
AvilaAndre: Didn't think it through, fixing it right now.

src/test/kotlin/pt/up/fe/ni/website/backend/BackendApplicationTests.kt
AvilaAndre: It was, found it weird too.

AvilaAndre: Changed it, I changed the link to href because I didn't really understand if I was supposed to delete it, either way, I can delete it quickly if so.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
```diff
@@ -0,0 +1,16 @@
...
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+
+@Entity
+class Post(
+    var title: String,
+    var body: String,
+    var href: String,
+    var templatePath: String,
+    var publishDate: Date,
```
BrunoRosendo:
```suggestion
    val publishDate: Date,
```

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Date
+import javax.persistence.Entity
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+
+@Entity
+class Post(
+    var title: String,
+    var body: String,
+    var href: String,
```
BrunoRosendo: I meant to remove it. I meant the `href` as an example of just creating links with HTML code, which is inside the body string

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
```diff
@@ -0,0 +1,16 @@
...
+    var templatePath: String,
```
BrunoRosendo:
```suggestion
    var thumbnailPath: String,
```
My bad, I think it was my typo in the other review

BrunoRosendo (changes requested):
Almost there

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
limwa: @BrunoRosendo I was under the impression that these were blog posts. If so, wouldn't it make sense to actually have an href in the model. For instance, I could create a blog post at https://ni.fe.up.pt/blog/how-we-make-lives-for-up-students-easier. In this case, the href would be the part after `blog/`. Are you sure it's best to remove it?

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
limwa: I think it might also be worth it to create a "lastUpdatedAt" date. Wdyt @BrunoRosendo @AvilaAndre?

limwa: I think it might also be worth it to create a "lastUpdatedAt" date. Wdyt @BrunoRosendo @AvilaAndre?

BrunoRosendo: I agree. Actually, publishDate and lastUpdatedAt should probably be automatically created by the BE, right? I completely missed that

BrunoRosendo: Check this thread https://stackoverflow.com/questions/51709727/spring-boot-jpacreateddate-lastmodifieddate-not-being-populated-when-saving-th

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
BrunoRosendo: Oh I was talking about something else. We're using the ID for the GET request. I agree that looks better and we could maybe query with that href instead of the ID. I would say create an issue for it. What do you think?

limwa: Yes, that sounds good!

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,36 @@
+package pt.up.fe.ni.website.backend.service
+
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.stereotype.Service
+import pt.up.fe.ni.website.backend.model.Post
+import pt.up.fe.ni.website.backend.repository.PostRepository
+
+class PostDTO {
```
BrunoRosendo:
```suggestion
class PostDto {
```

BrunoRosendo (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
BrunoRosendo: @AvilaAndre Maybe this post helps more https://self-learning-java-tutorial.blogspot.com/2019/08/spring-data-jpa-createddate.html

I think you can try adding @CreatedDate to the publishDate and create a lastUpdatedAt with @LastModifiedDate. This will hopefully do everything automatically but please test it. Also test if it's not possible to create a post with an arbitrary publishDate (it shouldn't be possible)

AvilaAndre: It worked! It isn't possible to create a Post with an arbitrary date, I think it is because it changes the publishDate before inserting it into the repository so it ends up changing the arbitrary value anyways. Updating it is also impossible as there is no publishDate on the PostDto. (I tested it)

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,36 @@
...
+    fun createPost(post: Post) = repository.save(post)
+
+    fun updatePost(PostId: Long, post: PostDto) {
+        val targetPost = repository.findByIdOrNull(PostId) ?: throw NoSuchElementException("Post Not Found")
+        targetPost.title = post.title ?: targetPost.title
+        targetPost.body = post.body ?: targetPost.body
+        targetPost.thumbnailPath = post.thumbnailPath ?: targetPost.thumbnailPath
+
+        repository.save(targetPost)
+    }
```
BrunoRosendo: We should return the edited post here
```suggestion
    fun updatePost(PostId: Long, post: PostDto): Post {
        val targetPost = repository.findByIdOrNull(PostId) ?: throw NoSuchElementException("Post Not Found")
        targetPost.title = post.title ?: targetPost.title
        targetPost.body = post.body ?: targetPost.body
        targetPost.thumbnailPath = post.thumbnailPath ?: targetPost.thumbnailPath

        return repository.save(targetPost)
    }
```

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -0,0 +1,36 @@
...
+    var body: String? = null
+    var thumbnailPath: String? = null
+}
+
+@Service
+class PostService(private val repository: PostRepository) {
+    fun getAllPosts(): List<Post> = repository.findAll().toList()
+
+    fun getPost(PostId: Long): Post =
+        repository.findByIdOrNull(PostId.toLong()) ?: throw NoSuchElementException("Post Not Found")
```
BrunoRosendo:
```suggestion
        repository.findByIdOrNull(PostId) ?: throw NoSuchElementException("Post Not Found")
```

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
BrunoRosendo: I think that, for consistency, we should return an empty object (`{}`) instead of an empty body. Changing the function return type to `Map<String, String>` and returning a `mapOf()` seems to do the trick

BrunoRosendo (changes requested):
The dates are working as expected! Just a few more things and we're all set

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
AvilaAndre: My bad, I thought I had deleted that already.

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
AvilaAndre: I committed the change where it returns the edited Post, I don't really understand what to do.

BrunoRosendo (approved):
Great job! :rocket:

PR #20, opened 2022-09-14
feat: add project endpoints

### LINKED ISSUE #4: projects: basic model and endpoints
Similarly to what we already have for events, we should have a simple first model for projects and 2/3 working endpoints to test it. This requires the creation of all the layers: model, repository, service and controller

### LINKED ISSUE #34: Fix `@RequestBody` vulnerability
The vulnerability can be summarized here: https://rules.sonarsource.com/java/tag/spring/RSPEC-4684

We could implement a proper DTO pattern or use an alternate solution. I'll leave it to discussion

limwa:
Closes #4.

This pull request adds basic project endpoints such as `Create New Project`, `Get All Projects`, `Get Project By Id`, `Delete Project By Id` and `Update Project By Id`.
I wasn't really imaginative when I wrote the Project model, so let me know if you think anything is missing.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -20,7 +20,7 @@ class ProjectController(private val service: ProjectService) {
     fun getAllProjects() = service.getAllProjects()
 
     @GetMapping("/{id}")
-    fun getProjectById(@PathVariable id: Long) = service.getProjectById(id).orElseThrow()!!
+    fun getProjectById(@PathVariable id: Long) = service.getProjectById(id)
 
     @PostMapping("/new")
     fun createNewProject(@RequestBody dto: ProjectDto): Project {
@@ -35,9 +35,5 @@ class ProjectController(private val service: ProjectService) {
     fun updatePostById(
         @PathVariable id: Long,
         @RequestBody dto: ProjectDto
-    ): Project {
-        val project = service.getProjectById(id).orElseThrow()!!
-        val newProject = dto.update(project)
-        return service.saveProject(newProject)
-    }
+    ) = service.updateProjectById(id, dto)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -4,25 +4,40 @@ import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import pt.up.fe.ni.website.backend.model.Project
+import pt.up.fe.ni.website.backend.model.dto.ProjectDto
 import pt.up.fe.ni.website.backend.service.ProjectService
 
 @RestController
 @RequestMapping("/projects")
 class ProjectController(private val service: ProjectService) {
 
-    @GetMapping("/")
+    @GetMapping
     fun getAllProjects() = service.getAllProjects()
-
-    @PostMapping("/new")
-    fun createNewProject(@RequestBody project: Project) = service.createProject(project)
 
     @GetMapping("/{id}")
     fun getProjectById(@PathVariable id: Long) = service.getProjectById(id).orElseThrow()!!
 
+    @PostMapping("/new")
+    fun createNewProject(@RequestBody dto: ProjectDto): Project {
+        val project = dto.update(Project.BLANK_PROJECT)
+        return service.saveProject(project)
+    }
+
     @DeleteMapping("/{id}")
     fun deleteProjectById(@PathVariable id: Long) = service.deleteProjectById(id)
+
+    @PutMapping("/{id}")
+    fun updatePostById(
+        @PathVariable id: Long,
+        @RequestBody dto: ProjectDto
+    ): Project {
+        val project = service.getProjectById(id).orElseThrow()!!
+        val newProject = dto.update(project)
+        return service.saveProject(newProject)
+    }
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -1,0 +1,28 @@
+package pt.up.fe.ni.website.backend.controller
+
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import pt.up.fe.ni.website.backend.model.Project
+import pt.up.fe.ni.website.backend.service.ProjectService
+
+@RestController
+@RequestMapping("/projects")
+class ProjectController(private val service: ProjectService) {
+
+    @GetMapping("/")
+    fun getAllProjects() = service.getAllProjects()
+
+    @PostMapping("/new")
+    fun createNewProject(@RequestBody project: Project) = service.createProject(project)
+
+    @GetMapping("/{id}")
+    fun getProjectById(@PathVariable id: Long) = service.getProjectById(id).orElseThrow()!!
+
+    @DeleteMapping("/{id}")
+    fun deleteProjectById(@PathVariable id: Long) = service.deleteProjectById(id)
+}
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -24,7 +24,7 @@ class ProjectController(private val service: ProjectService) {
 
     @PostMapping("/new")
     fun createNewProject(@RequestBody dto: ProjectDto): Project {
-        val project = dto.update(Project.BLANK_PROJECT)
+        val project = dto.create()
         return service.saveProject(project)
     }
 
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
@@ -6,8 +6,13 @@ import javax.persistence.Id
 
 @Entity
 class Project(
-    val name: String,
-    val description: String,
+    var name: String,
+    var description: String,
     @Id @GeneratedValue
-    val id: Long? = null,
-)
+    val id: Long? = null
+) {
+    companion object {
+        @JvmStatic
+        val BLANK_PROJECT = Project("", "")
+    }
+}
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
@@ -10,9 +10,4 @@ class Project(
     var description: String,
     @Id @GeneratedValue
     val id: Long? = null
-) {
-    companion object {
-        @JvmStatic
-        val BLANK_PROJECT = Project("", "")
-    }
-}
+)
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
@@ -1,0 +1,13 @@
+package pt.up.fe.ni.website.backend.model
+
+import javax.persistence.Entity
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+
+@Entity
+class Project(
+    val name: String,
+    val description: String,
+    @Id @GeneratedValue
+    val id: Long? = null,
+)
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
@@ -1,0 +1,26 @@
+package pt.up.fe.ni.website.backend.model.dto
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import javax.validation.ConstraintViolationException
+import javax.validation.Validation
+
+open class Dto<T : Any> {
+
+    companion object {
+        private val objectMapper = ObjectMapper()
+        private val validatorFactory = Validation.buildDefaultValidatorFactory()
+    }
+
+    fun update(entity: T): T {
+        val newEntity = objectMapper.updateValue(entity, this)
+
+        val validator = validatorFactory.validator
+        val violations = validator.validate(newEntity)
+
+        if (violations.isNotEmpty()) {
+            throw ConstraintViolationException(violations)
+        }
+
+        return newEntity
+    }
+}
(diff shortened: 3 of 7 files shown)


BrunoRosendo:
I haven't reviewed but it might be a good idea to take a look at #19, since everything there probably applies here too

limwa:
Okay, will do! Thanks!

BrunoRosendo (changes requested):
This is looking good but maybe we should already implement the patch endpoint.
Additionally, for consistency, we should treat exceptions (among other things) the same way as #19. You can maybe go there and give your opinion. After that's merged, we can follow it for all other controllers/services

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -0,0 +1,28 @@
...
+
+@RestController
+@RequestMapping("/projects")
+class ProjectController(private val service: ProjectService) {
+
+    @GetMapping("/")
+    fun getAllProjects() = service.getAllProjects()
+
+    @PostMapping("/new")
+    fun createNewProject(@RequestBody project: Project) = service.createProject(project)
```
DoStini: https://rules.sonarsource.com/java/tag/spring/RSPEC-4684

DoStini: Found this online

limwa: Thank you! We should open issues to fix these vulnerabilities!

BrunoRosendo: That is a very nice point. Maybe we should implement the proper DTO pattern in our project. It felt over-engineered to me but maybe it's worth it

limwa: I'm almost finishing my DTO implementation. The only part that is missing is saving the result to the DB. I'll push it once it's ready!

limwa: I've installed SonarLint on IntellIJ and it didn't detect this vulnerability because we are using Kotlin :upside_down_face: 
Looks like we'll just have to be very careful...

BrunoRosendo:
@limwa Is this ready? Also, this seems to be related to #34 so I'll add it here

limwa:
> @limwa Is this ready? Also, this seems to be related to #34 so I'll add it here

Yes, it's ready!
Regarding the fix of #34, it shouldn't be closed right now because the other controllers still need fixing!

BrunoRosendo:
> > @limwa Is this ready? Also, this seems to be related to #34 so I'll add it here
> 
> Yes, it's ready! Regarding the fix of #34, it shouldn't be closed right now because the other controllers still need fixing!

Don't worry, we'll open it again. This way we can open the issue and see all the PRs contributing to it

src/main/kotlin/pt/up/fe/ni/website/backend/BackendApplication.kt
```diff
@@ -3,7 +3,9 @@ package pt.up.fe.ni.website.backend
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
 import org.springframework.data.jpa.repository.config.EnableJpaAuditing
+import org.springframework.web.servlet.config.annotation.EnableWebMvc
 
+@EnableWebMvc
```
BrunoRosendo: Why is this necessary now? (just curious)

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -0,0 +1,43 @@
...
+
+@RestController
+@RequestMapping("/projects")
+class ProjectController(private val service: ProjectService) {
+
+    @GetMapping
+    fun getAllProjects() = service.getAllProjects()
+
+    @GetMapping("/{id}")
+    fun getProjectById(@PathVariable id: Long) = service.getProjectById(id).orElseThrow()!!
```
BrunoRosendo: We should use the same logic as the post endpoint. Could you check it and give your thoughts?

src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
```diff
@@ -0,0 +1,21 @@
...
+class Project(
+    var name: String,
+    var description: String,
+    @Id @GeneratedValue
+    val id: Long? = null
+) {
+    companion object {
+        @JvmStatic
+        val BLANK_PROJECT = Project("", "")
+    }
```
BrunoRosendo: This is the only part I don't really like about this implementation. I think a `create` function in `Dto` would be a better abstraction, it seems weird to me to have this in the model and having to use it in the controller.

If the problem is you don't have a `createValue` the same way you have an `updateValue` (I didn't find it but I didn't look too much into it) then I think the `ModelMapper` could be a better option. If you decide to not use it then here's a solution from the top of my head:
- Create an abstract function in `Dto` called `blankEntity()`
- Create a wrapper `fun create() = update(blankEntity())`
- Override `blankEntity()` in `PostDto` instead of adding it to the model

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    @DeleteMapping("/{id}")
+    fun deleteProjectById(@PathVariable id: Long) = service.deleteProjectById(id)
+
+    @PutMapping("/{id}")
+    fun updatePostById(
+        @PathVariable id: Long,
+        @RequestBody dto: ProjectDto
+    ): Project {
+        val project = service.getProjectById(id).orElseThrow()!!
```
BrunoRosendo: Same thing with treating exceptions the same way as in `Post`

BrunoRosendo (changes requested):
Let me start by saying I like this implementation of the DTO pattern, it feels really clean. I think you were right about it and we should use it going forward (the PUTs too).

Did you check [this library](http://modelmapper.org/getting-started/)? ([examples](https://www.baeldung.com/entity-to-and-from-dto-for-a-java-spring-application)). I think it does what you're trying to do here and it might result in an even better solution. However, I don't have a problem with your implementation besides my comment below about `create`. If you find a solution for that then I understand if you prefer not to add dependencies.

Let me hear what you think. Good job so far!

BrunoRosendo:
This will also need a rebase

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    fun update(entity: T): T {
+        val newEntity = objectMapper.updateValue(entity, this)
+
+        val validator = validatorFactory.validator
+        val violations = validator.validate(newEntity)
+
+        if (violations.isNotEmpty()) {
+            throw ConstraintViolationException(violations)
```
BrunoRosendo, line 43: This exception will not work with current `ErrorController`. Hopefully we only need to update the exception handled there.

ModelMapper uses `ValidationException` so watch out for that if you use it

BrunoRosendo:
One more thing. Please add the `dto` package to the structure setion of the README. We have a lot of layers so we need to keep some documentation

src/main/kotlin/pt/up/fe/ni/website/backend/BackendApplication.kt
limwa: I don't know, I just started spamming Annotations when I was testing custom validations. Probably isn't needed anymore...

BrunoRosendo (commented):
The new create looks great. It's a shame `reified` doesn't work for classes but this is already good

limwa:
@BrunoRosendo the PR is ready for review!
I haven't touched the error controller (so that ConstraintValidationExceptions get picked up by it), but let me know if you want me to make that work or if we should leave it to another issue!

BrunoRosendo:
> @BrunoRosendo the PR is ready for review! I haven't touched the error controller (so that ConstraintValidationExceptions get picked up by it), but let me know if you want me to make that work or if we should leave it to another issue!

It's ok, we can do it in #34 . I may pick that up

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
```diff
@@ -0,0 +1,77 @@
...
+open class Dto<T : Any> {
+
+    @JsonIgnore
+    private val entityClass: KClass<T>
+
+    constructor() : this(null)
+
+    constructor(conversionClass: KClass<T>?) {
+        this.entityClass = DtoReflectionUtils.getTypeConversionClassWithCache(this::class, conversionClass)
+    }
```
BrunoRosendo, line 21:
```suggestion
    constructor(conversionClass: KClass<T>? = null) {
        this.entityClass = DtoReflectionUtils.getTypeConversionClassWithCache(this::class, conversionClass)
    }
```
Am I wrong to assume this is the same? Seems simpler to me

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    companion object {
+        private val objectMapper = ObjectMapper()
+        private val validatorFactory = Validation.buildDefaultValidatorFactory()
+    }
```
BrunoRosendo, line 26: I'm curious about `fields vs companion object`. Are there situations when one is preferred over the other? I'm asking for my own code too

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        private fun <T : Any> ensureEntity(clazz: KClass<T>?): KClass<T> {
+            if (clazz == null || !clazz.hasAnnotation<Entity>()) {
+                throw IllegalArgumentException("$clazz is not a DTO of an entity")
+            }
+
+            return clazz
+        }
+    }
+}
```
BrunoRosendo: I assume you tested this and it looks good at first glance. However, the code is pretty complex so it'd be nice to have comments explain things better. I'll leave some questions that crossed my mind when I was reading through the code:
- Why is `@Suppress("UNCHECKED_CAST")` needed?
- Why are the types cached instead of always calculating them? Is there a performance hit without cache?
- How does the `jvmErasure` work?
- What is the goal behind `fallback` in `getTypeConversionClassWithCache`?

Comments explaining these things would be better, imo

BrunoRosendo (changes requested):
Good job. You put some work into this and the DTO system is looking well!
I left some comments that are mostly cosmetic and also some doubts

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
limwa: At runtime, it is the same, but for IDEs it's not. The current implementation shows two possible constructors, one with no arguments and one with a class for an argument. The suggested implementation only shows one option. Since the purpose of the class argument was a bit confusing already, I chose the two constructors option, but do let me know if you want me to change it.

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
limwa: A companion object is the equivalent of a static class, that is a class that only exists once for all instances of Dto (you can think of it as putting static fields on Dto). That means that there will only be one validator factory, one object mapper and one conversion type cache across all Dto. In the case of ObjectMapper and Validator Factory, to reduce the number of those we have in our code, I've put them in the companion object. In the case of the type conversion cache, it has to be in the companion object since that information needs to be shared across instances of Dto.

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
limwa: `@Suppress("UNCHECKED_CAST")` is a compiler hint to suppress all warnings in that method which are related to unchecked casts.

When you cast an object to a string, the JVM checks if that object really is a String before proceeding. If it's not, it'll throw a cast exception. That's the reason why you should always use `instanceof` before any cast.

But what about `List`? What happens if I want to convert a `List<Integer>` to a `List<String>`? You can't really cast a `String` to an `Integer`, but generics don't really exist at runtime, so, in practice, you'd be converting from `List` to `List`, which seems valid to the JVM. That's an unchecked cast.

limwa: Yes, there is a performance hit when using reflection. Reflective operations are take a lot of time since they can't be optimized by javac at compile-time. As a rule of thumb, you should minimize the amount of reflective operations in your code, and since the class argument stays the same for every different kind of Dto, we can cache it!

Right now, after reading your comment, I'm evaluating if we should use https://jmapper-framework.github.io/jmapper-core/, which completely avoids reflection by automatically generating bytecode that the JVM can execute as if it were normal, native code.

limwa: To be completely honest, I have no idea of how `jvmErasure` works. It's a Kotlin Reflect feature that seems to completely recover generic type information at runtime. I use it to determine the class of T, the type argument of Dto.

limwa: The goal behind fallback is to reuse code. We have a type cache and *all* types in the cache are verified to be types that are annotated with `@Entity`. Therefore, with fallback, we skip the automatic type determination but we still ensure it's an entity before putting it on the cache.

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
BrunoRosendo: Got it, it's ok like this

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt
BrunoRosendo: I'll leave the decision up to you. I think the implementation with the type in the constructor was already simple and ok. This one seems good too given the cache will take care of performance.

BrunoRosendo: Thanks for the explanations. Please write some comments in the code if you decide to stick with this!

src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
```diff
@@ -0,0 +1,16 @@
+package pt.up.fe.ni.website.backend.model
+
+import org.springframework.data.jpa.domain.support.AuditingEntityListener
+import javax.persistence.Entity
+import javax.persistence.EntityListeners
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+
+@Entity
+@EntityListeners(AuditingEntityListener::class)
```
BrunoRosendo: What is this for?

BrunoRosendo (commented):
Sorry for being annoying, I got one more question xd

src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
BrunoRosendo: This was used for `publishDate` and `lastUpdatedAt` in the Post. Do we need it here?

limwa: No xD

limwa: Thanks!

BrunoRosendo:
@limwa I was testing this and I can't create new projects. It seems to be related to the project's ID when converting from the DTO

BrunoRosendo (approved):
Approving now, this will later be completed/refactored. Can you please open an issue for it @limwa ?
I was not expecting such a long documentation but it's ok for now 😅

limwa:
Yes, I'll take care of that later

PR #22, opened 2022-09-17
Validation and Error Handling

### LINKED ISSUE #7: events: implement middleware for validations
The main job of this issue is to investigate and find the best way to implement middleware in Spring Boot. We need a clean, consistent and scalable way of doing it for all the endpoints.

After that is done, you can simply create an example for the `events/new` endpoint that validates something simple such as a valid date or title with a maximum number of characters

### LINKED ISSUE #15: errors: add global error handling

BrunoRosendo:
closes #7 
closes #15 

This PR adds simple validation for the `events` endpoints, serving as example for the other resources. It also introduces error handling with a consistent format of error messages, including validation and other types of errors.

Errors around Jackson and its binding errors are not working as great as I'd want them to. Since I spent way too long hitting my head with this problem, I created #27 and left that problem for later.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
@@ -1,14 +1,18 @@
 package pt.up.fe.ni.website.backend.controller
 
+import com.fasterxml.jackson.databind.exc.InvalidFormatException
+import com.fasterxml.jackson.module.kotlin.MissingKotlinParameterException
 import org.springframework.boot.web.servlet.error.ErrorController
 import org.springframework.http.HttpStatus
+import org.springframework.http.converter.HttpMessageNotReadableException
 import org.springframework.validation.FieldError
 import org.springframework.validation.ObjectError
 import org.springframework.web.bind.MethodArgumentNotValidException
 import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.bind.annotation.RestControllerAdvice
-import org.springframework.web.servlet.NoHandlerFoundException
 
 data class SimpleError(
     val message: String,
@@ -18,10 +22,11 @@ data class SimpleError(
 
 data class CustomError(val errors: List<SimpleError>)
 
+@RestController
 @RestControllerAdvice
 class ErrorController : ErrorController {
 
-    @ExceptionHandler(NoHandlerFoundException::class)
+    @RequestMapping("/**")
     @ResponseStatus(HttpStatus.NOT_FOUND)
     fun endpointNotFound(): CustomError = wrapSimpleError("invalid endpoint")
 
@@ -42,6 +47,35 @@ class ErrorController : ErrorController {
         return CustomError(errors)
     }
 
+    @ExceptionHandler(HttpMessageNotReadableException::class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    fun invalidRequestBody(e: HttpMessageNotReadableException): CustomError {
+        when (val cause = e.cause) {
+            is InvalidFormatException -> {
+                val type = cause.targetType.simpleName.lowercase()
+                return wrapSimpleError(
+                    "must be $type",
+                    value = cause.value
+                )
+            }
+
+            is MissingKotlinParameterException -> {
+                return wrapSimpleError(
+                    "required",
+                    param = cause.parameter.name
+                )
+            }
+        }
+
+        return wrapSimpleError(e.message ?: "invalid request body")
+    }
+
+    @ExceptionHandler(NoSuchElementException::class)
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    fun elementNotFound(e: NoSuchElementException): CustomError {
+        return wrapSimpleError(e.message ?: "element not found")
+    }
+
     @ExceptionHandler(Exception::class)
     @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
     fun unexpectedError(e: Exception): CustomError {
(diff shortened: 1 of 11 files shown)


src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
```diff
@@ -1,13 +1,89 @@
...
+    @ExceptionHandler(NoSuchElementException::class)
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    fun elementNotFound(e: NoSuchElementException): CustomError {
+        return wrapSimpleError(e.message ?: "element not found")
+    }
+
+    @ExceptionHandler(Exception::class)
+    @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
+    fun unexpectedError(e: Exception): CustomError {
+        System.err.println(e)
```
limwa, line 82: We should create an issue to use proper logging instead of stdout and stderr

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -1,15 +1,26 @@
 package pt.up.fe.ni.website.backend.model
 
+import com.fasterxml.jackson.annotation.JsonProperty
 import java.util.Date
 import javax.persistence.Entity
 import javax.persistence.GeneratedValue
 import javax.persistence.Id
+import javax.validation.constraints.Size
+import pt.up.fe.ni.website.backend.model.constants.EventConstants as constants
```
limwa:
```suggestion
import pt.up.fe.ni.website.backend.model.constants.EventConstants as Constants
```
I would change it to start with a capital letter. This way, it wouldn't be confused with a variable.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -1,15 +1,26 @@
...
 
 @Entity
 class Event(
+    @JsonProperty(required = true)
```
limwa, line 13: Why are these annotations necessary?

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
```diff
@@ -8,14 +9,42 @@ import javax.persistence.Entity
 import javax.persistence.EntityListeners
 import javax.persistence.GeneratedValue
 import javax.persistence.Id
+import javax.validation.constraints.NotEmpty
+import javax.validation.constraints.Size
+import pt.up.fe.ni.website.backend.model.constants.PostConstants as constants
```
limwa: Same comment as the last one, on the event file

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -5,23 +5,18 @@ import org.springframework.stereotype.Service
...
     fun getPost(PostId: Long): Post =
-        repository.findByIdOrNull(PostId) ?: throw NoSuchElementException("Post Not Found")
+        repository.findByIdOrNull(PostId) ?: throw NoSuchElementException("post not found with id $PostId")
 
-    fun createPost(post: Post) = repository.save(post)
+    fun createPost(post: Post): Post = repository.save(post)
 
-    fun updatePost(PostId: Long, post: PostDto): Post {
-        val targetPost = repository.findByIdOrNull(PostId) ?: throw NoSuchElementException("Post Not Found")
+    fun updatePost(PostId: Long, post: Post.PatchModel): Post {
```
limwa:
```suggestion
    fun updatePost(postId: Long, post: Post.PatchModel): Post {
```
Since you've changed this line already, can you please fix this?

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -30,7 +25,7 @@ class PostService(private val repository: PostRepository) {
     }
 
     fun deletePost(PostId: Long): Map<String, String> {
```
limwa:
```suggestion
    fun deletePost(postId: Long): Map<String, String> {
```
Same here

limwa (changes requested):
So far, it's looking good. I'm not particularly fond of having the PatchModel as an inner class, but if the others don't mind, it's fine by me.

BrunoRosendo:
> So far, it's looking good. I'm not particularly fond of having the PatchModel as an inner class, but if the others don't mind, it's fine by me.

I don't think it's perfect either... I wanted to somehow bind it with the model to have validation all in the same place, but I'm definitely taking suggestions regarding this

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
limwa: To clarify, I'm not saying they aren't needed. I'm just curious about the reason so I can be more careful in my own PRs.

BrunoRosendo: This annotation makes no difference here. However, I was playing with numerical fields and the annotation is needed in those, otherwise Jackson uses the default value instead of throwing an error. I assume this happens for every class that is mapped to a primitive in Java. I decided to put the annotations everywhere for consistency

limwa: Thanks!

limwa (approved):
LGTM! 🚀

Even though the code has vulnerabilities present, as we discovered recently, to fix them falls out of the scope of this PR. We'll have to fix them ASAP though.

Good job! 😁

PR #46, opened 2022-11-08
Models and endpoints for user accounts

### LINKED ISSUE #3: accounts: implement basic model and roles
This issue is meant to have an investigation part where you need to discover the best way of implementing accounts with roles and permissions in Spring Boot, this is a very common need in APIs so I'm sure the framework will have something to help us here.

After that, you should implement an account model with basic functionalities and the necessary roles. You can take a look at [the accounts of the current website](https://github.com/NIAEFEUP/Website-NIAEFEUP/blob/develop/src/models/User.js). The point is just for inspiration and not to recreate their model, we can have something different that fits our needs here.

## Tasks
- [x] Create an Account model
- [x] Create basic Account endpoints
- [x] Create models for Roles and Permissions
- [x] Integrate the Roles with the Account model

BrunoRosendo:
closes #3 

Currently missing tests and a fix regarding relationship between Account and CustomWebsite

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
@@ -30,7 +30,7 @@ class Account(
     @field:Email
     var email: String,
 
-    @field:Size(min = Constants.Name.minSize, max = Constants.Name.maxSize)
+    @field:Size(min = Constants.Bio.minSize, max = Constants.Bio.maxSize)
     var bio: String?,
 
     @field:Past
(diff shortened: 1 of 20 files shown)


src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
```diff
@@ -0,0 +1,60 @@
...
+    @field:NullOrNotBlank
+    @field:URL
+    var linkedin: String?,
+
+    @field:NullOrNotBlank
+    @field:URL
+    var github: String?,
+
+    @JoinColumn
+    @OneToMany(cascade = [CascadeType.ALL], fetch = FetchType.EAGER)
```
BrunoRosendo, line 53: Details about these annotations' behaviour can be found at:
https://stackoverflow.com/questions/52887653/onetomany-relationship-does-not-save-the-primary-key-of-the-parent-in-the-child
https://stackoverflow.com/questions/2990799/difference-between-fetchtype-lazy-and-eager-in-java-persistence-api

BrunoRosendo:
This PR is only missing tests now

ruialves35 (approved):

PR #47, opened 2022-11-10
Adds error message templates

### LINKED ISSUE #29: errors: store error messages as constants
Currently, error messages are hard coded and identical errors are repeated across the code. It would be better to have them stored in an object (the same way as `Event` and `Post` have their constants).

However, there's a catch. Many of the error messages need to use a template string. Passing them to a function doesn't work because they're not constant at compile time, so we need to find an alternative solution.

jamcunha:
Closes #29

Couldn't store error messages that were used in annotations (in models).
This PR adds an object that contains functions to create a error messages for services such as `postNotFound` and `projectNotFound`.

# Review checklist


limwa:
I think that we could eliminate all duplicate annotations by using a resource bundle (.properties file under src/main/resources). It can take parameters such as min and max and, that way, we could use the same message for all annotations of that type. Could you please have a look at it? @BrunoRosendo what do you think?

BrunoRosendo:
> I think that we could eliminate all duplicate annotations by using a resource bundle (.properties file under src/main/resources). It can take parameters such as min and max and, that way, we could use the same message for all annotations of that type. Could you please have a look at it? @BrunoRosendo what do you think?

Can you show an example of that? I found this article but it doesn't mention how these resources could take parameters
https://www.baeldung.com/spring-custom-validation-message-source

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -11,11 +11,19 @@ import pt.up.fe.ni.website.backend.model.constants.EventConstants as Constants
 @Entity
 class Event(
     @JsonProperty(required = true)
-    @field:Size(min = Constants.Title.minSize, max = Constants.Title.maxSize)
+    @field:Size(
+        min = Constants.Title.minSize,
+        max = Constants.Title.maxSize,
+        message = "size must be between ${Constants.Title.minSize} and ${Constants.Title.maxSize}"
+    )
```
BrunoRosendo: I don't think we need to override the `Size`'s message, since the default already works well for us. What do you think?

This also applies to the other models

BrunoRosendo (changes requested):
Good job, just a small change request for now.

Let's also discuss the option of using resource bundles for this. It might be a viable option because it's used in the javax's validation itself and it would allow us to more easily have translations in the future, if needed

limwa:
https://stackoverflow.com/a/70936114/14838815 @BrunoRosendo @jamcunha I've been looking into resource bundles and I found this which describes exactly what I was thinking. This way, we can easily change any validation message across the entire backend with a single line change!

BrunoRosendo:
> 

I think this could be a good alternative. @jamcunha Do you want to give it a try?

BrunoRosendo:
The problem seems to be related to our `Dto` class, in the following line: https://github.com/NIAEFEUP/website-niaefeup-backend/blob/02430c1611dced5a9ce06229c9fe6e59616c0b44/src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/Dto.kt#L25

I was trying to replace this initialization with Spring Boot's dependency injection, by using `@Autowired`. However, `Dto` would need to be a component and a component apparently can't be extended by other Dtos. I will try to look for a solution later but please explore the problem before me if you can

limwa:
@BrunoRosendo regarding the use of .properties files it should be noted that no conditions can be used inside the strings. Using things such as "name must have size greater than 4", instead of "name must have size greater than 4 and less than 2147483647", is not possible (without changing the annotation's properties. Either way, I think this is a good thing as, realistically, we should be imposing a hard limit, such as 1000, 10000 or even 100000.

src/main/kotlin/pt/up/fe/ni/website/backend/config/validation/ValidationConfig.kt
```diff
@@ -0,0 +1,29 @@
...
+        val validator = LocalValidatorFactoryBean()
+        validator.setValidationMessageSource(messageSource)
+        return validator
+    }
+
+    @Bean
+    fun messageSource(): MessageSource {
+        val bean = ReloadableResourceBundleMessageSource()
+        bean.addBasenames(
+            "classpath:pt.up.fe.ni.website.backend.model.validator.MessageValidator",
```
BrunoRosendo:
```suggestion
```
I don't think this is needed and the classpath is outdated

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
```diff
@@ -0,0 +1,11 @@
...
+
+object ErrorMessages {
+    fun postNotFound(postId: Long): String {
+        return "post not found with id $postId"
+    }
+
+    fun projectNotFound(id: Long): String {
+        return "project not found with id $id"
+    }
+}
```
BrunoRosendo: This is already pretty cool but I wonder if we can move every error message to the resources. What do you think?

BrunoRosendo (changes requested):
Good job so far! I left some suggestions

BrunoRosendo:
@jamcunha Can we use the resources for the validation message in `NullOrNotBlank`? I believe we can

src/main/resources/validation_errors.properties
```diff
@@ -0,0 +1,2 @@
+size.min = size must be greater or equal to {min}
+null_blank.error = must be null or not blank
```
BrunoRosendo:
```suggestion
null_or_not_blank.error = must be null or not blank
```
It's better to be explicit, especially when working with constants

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
```diff
@@ -0,0 +1,35 @@
...
+        return "invalid credentials"
+    }
+
+    fun invalidRefreshToken(): String {
+        return "invalid refresh token"
+    }
+
+    fun expiredRefreshToken(): String {
+        return "refresh token has expired"
+    }
```
BrunoRosendo:
```suggestion
   fun postNotFound(postId: Long) = "post not found with id $postId"

    fun projectNotFound(id: Long) = "project not found with id $id"

    fun emailAlreadyExists() = "email already exists"

    fun accountNotFound(id: Long) = "account not found with id $id"

    fun emailNotFound(email: String) = "account not found with email $email"

    fun invalidCredentials() = "invalid credentials"

    fun invalidRefreshToken() = "invalid refresh token"

    fun expiredRefreshToken() = "refresh token has expired"
```
I think this is simpler and easier to read

BrunoRosendo:
```suggestion
    const val emailAlreadyExists = "email already exists"

    const val invalidCredentials = "invalid credentials"

    const val invalidRefreshToke = "invalid refresh token"

    const val expiredRefreshToken = "refresh token has expired"

    fun postNotFound(postId: Long) = "post not found with id $postId"

    fun projectNotFound(id: Long) = "project not found with id $id"

    fun accountNotFound(id: Long) = "account not found with id $id"

    fun emailNotFound(email: String) = "account not found with email $email"
```

My second suggestion is to change the constant error messages to `const val` instead of functions, since they always return a static value. But I'll leave that decision up to you

BrunoRosendo (changes requested):
Good job, the implementation is done and working as expected! My last comments are just some minor cosmetic suggestions we can do by using some Kotlin features like type inference.

BrunoRosendo (approved):
Nice!

Issue #42, opened 2022-11-07
accounts: implement generations


BrunoRosendo:
In order to have a history of NIAEFEUP's members over the years, we should support generations. We should be able to recall the roles of the members as well (member, recruit, president, etc.).

It's possible to see a developed UML chart in #3 regarding this feature

Issue #41, opened 2022-11-07
projects/events: add team members


BrunoRosendo:
The projects and the events (the "Item" general class created in #23) should contain teams of their current members or organizers. We should:
- [x] Add a list of members to the Item's model
- [x] Create endpoints capable of adding and removing team members

PR #49, opened 2022-11-15
Feature/add permissions and roles

### LINKED ISSUE #3: accounts: implement basic model and roles

limwa:
Closes #3.

This pull request adds permissions and roles.
Additionally, it adds roles to accounts. 

This PR will most likely not contain any controllers and that will be done in a later PR.

# Review checklist
-   [x] Properly documents API changes in `docs/openapi.yml`


limwa:
@BrunoRosendo I'm getting a weird error when testing

```
org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: failed to lazily initialize a collection of role: pt.up.fe.ni.website.backend.model.Account.roles: could not initialize proxy - no Session
```

Do you know what could be causing this error?

BrunoRosendo:
> @BrunoRosendo I'm getting a weird error when testing
> 
> ```
> org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: failed to lazily initialize a collection of role: pt.up.fe.ni.website.backend.model.Account.roles: could not initialize proxy - no Session
> ```
> 
> Do you know what could be causing this error?

I'm on the phone and I'll give you a detailed answer tomorrow. But I'm pretty sure you need to set the field as Eager, check account and CustomWebsite

limwa:
> > @BrunoRosendo I'm getting a weird error when testing
> > ```
> > org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: failed to lazily initialize a collection of role: pt.up.fe.ni.website.backend.model.Account.roles: could not initialize proxy - no Session
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you know what could be causing this error?
> 
> I'm on the phone and I'll give you a detailed answer tomorrow. But I'm pretty sure you need to set the field as Eager, check account and CustomWebsite

Thanks!

BrunoRosendo:
> > @BrunoRosendo I'm getting a weird error when testing
> > ```
> > org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: failed to lazily initialize a collection of role: pt.up.fe.ni.website.backend.model.Account.roles: could not initialize proxy - no Session
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you know what could be causing this error?
> 
> I'm on the phone and I'll give you a detailed answer tomorrow. But I'm pretty sure you need to set the field as Eager, check account and CustomWebsite

Adding on this, the default behaviour in relationships is to load lazily, which excludes serialization. That makes sense since you wouldn't want to retrieve all the relationships from the DB and that would most likely cause circular dependencies, so we have to explicitly set the fetch type as eager if we want to couple two tables together when fetching data

limwa:
> > > @BrunoRosendo I'm getting a weird error when testing
> > > ```
> > > org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: failed to lazily initialize a collection of role: pt.up.fe.ni.website.backend.model.Account.roles: could not initialize proxy - no Session
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Do you know what could be causing this error?
> > 
> > I'm on the phone and I'll give you a detailed answer tomorrow. But I'm pretty sure you need to set the field as Eager, check account and CustomWebsite
> 
> Adding on this, the default behaviour in relationships is to load lazily, which excludes serialization. That makes sense since you wouldn't want to retrieve all the relationships from the DB and that would most likely cause circular dependencies, so we have to explicitly set the fetch type as eager if we want to couple two tables together when fetching data

Then, do you know how we can still fetch lazily and stop serialization? For instance, when returning a role, I don't want to return all of the accounts that belong to the role

BrunoRosendo:
> > > > @BrunoRosendo I'm getting a weird error when testing
> > > > ```
> > > > org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: failed to lazily initialize a collection of role: pt.up.fe.ni.website.backend.model.Account.roles: could not initialize proxy - no Session
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Do you know what could be causing this error?
> > > 
> > > 
> > > I'm on the phone and I'll give you a detailed answer tomorrow. But I'm pretty sure you need to set the field as Eager, check account and CustomWebsite
> > 
> > 
> > Adding on this, the default behaviour in relationships is to load lazily, which excludes serialization. That makes sense since you wouldn't want to retrieve all the relationships from the DB and that would most likely cause circular dependencies, so we have to explicitly set the fetch type as eager if we want to couple two tables together when fetching data
> 
> Then, do you know how we can still fetch lazily and stop serialization? For instance, when returning a role, I don't want to return all of the accounts that belong to the role

I haven't looked at the code yet but it should go like this:
- In the `Account`, fetch eagerly which includes all the roles.
- In the `Role`, fetch lazily (default) which does not include the associated accounts.

Edit: I realize I wasn't clear before, the tables don't become coupled when you add Eager fetching in the account, it's one-directional

src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
```diff
@@ -56,6 +58,24 @@ class Account(
...
+            result.addAll(role.permissions)
+
+            for (perActivity in role.perActivities) {
+                if (perActivity.activity == activity) {
+                    result.addAll(perActivity.permissions)
+                }
+            }
+        }
+    }
+}
```
BrunoRosendo: I think this is missing the return value. I'm also suggesting a functional version using lambdas that I think looks simpler. I know some people might not be as used to Kotlin's syntax so choose whatever you prefer or even a mix of both

```kotlin
fun getEffectivePermissionsForActivity(activity: Activity) = Permissions().apply {
        roles.forEach { role ->
            addAll(role.permissions)

            role.perActivities
                .find { it.activity == activity }
                ?.permissions
                ?.let { addAll(it) }
        }
    }
```
There's also this short version but I think it's too confusing

```kotlin
fun getEffectivePermissionsForActivity(activity: Activity) = roles
    .flatMap { role -> role.permissions + role.perActivities.find { it.activity == activity }?.permissions.orEmpty() }
```

BrunoRosendo: I might even prefer a version with the var declaration, lambdas and return. Many choices

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
```diff
@@ -21,6 +26,10 @@ abstract class Activity(
     @field:Size(min = Constants.Description.minSize, max = Constants.Description.maxSize)
     open val description: String,
 
+    @JoinColumn
+    @OneToMany(cascade = [CascadeType.ALL], fetch = FetchType.EAGER)
+    open val perRoles: List<@Valid PerActivityRole>,
```
BrunoRosendo: This will probably conflict with #41, we should discuss how to handle team members and the respective roles

BrunoRosendo: Also, what do you think about `activityRoles` or simply `roles`? I'm finding `perRoles` hard to grasp

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
```diff
@@ -0,0 +1,40 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToMany
+import jakarta.persistence.OneToMany
+import jakarta.validation.Valid
+import pt.up.fe.ni.website.backend.permissions.Permissions
+import pt.up.fe.ni.website.backend.permissions.PermissionsConverter
+
+@Entity
+class Role(
```
BrunoRosendo, line 17: We'll need to think about what should be fetched together with roles and other models. That can be decided in the issue where the endpoints are created, I'll create it

src/main/kotlin/pt/up/fe/ni/website/backend/permissions/Permission.kt
```diff
@@ -0,0 +1,7 @@
+package pt.up.fe.ni.website.backend.permissions
+
+enum class Permission(val bit: Int) {
```
BrunoRosendo: Maybe this `permissions/` package could go under `model/`? We're starting to have too many top level packages

src/main/kotlin/pt/up/fe/ni/website/backend/permissions/Permissions.kt
```diff
@@ -0,0 +1,25 @@
+package pt.up.fe.ni.website.backend.permissions
+
+import java.util.Collections
+import java.util.TreeSet
+
+class Permissions(private val permissions: Set<Permission> = Collections.emptySet()) : MutableSet<Permission> by TreeSet(permissions) {
```
BrunoRosendo:
```suggestion
class Permissions(
    private val permissions: Set<Permission> = Collections.emptySet()
) : MutableSet<Permission> by TreeSet(permissions) {
```
We might want to add line length limit to ktlint. Anyway, this change helps me read it

src/main/kotlin/pt/up/fe/ni/website/backend/permissions/Permissions.kt
```diff
@@ -0,0 +1,25 @@
...
+            val result = Permissions()
+            for (perm in Permission.values()) {
+                val encodedBit = encoded ushr perm.bit and 1L
+                if (encodedBit == 1L) {
+                    result.add(perm)
+                }
+            }
+
+            return result
+        }
```
BrunoRosendo: This can have a similar refactor if you wish. I didn't execute it though
```kotlin
fun fromLong(encoded: Long) = Permission.values()
    .filter { encoded ushr it.bit and 1L == 1L }
    .toSet()
    .let(::Permissions)
```

BrunoRosendo: I just think we should add a comment explaining how the shift works

src/main/kotlin/pt/up/fe/ni/website/backend/permissions/Permissions.kt
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    fun toLong() = permissions.fold(0L) { acc, perm ->
+        return acc or (1L shl perm.bit)
```
BrunoRosendo: Same thing about the comment

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,58 @@
...
+import pt.up.fe.ni.website.backend.model.Activity
+import pt.up.fe.ni.website.backend.model.PerActivityRole
+import pt.up.fe.ni.website.backend.model.Role
+import pt.up.fe.ni.website.backend.permissions.Permission
+import pt.up.fe.ni.website.backend.permissions.Permissions
+import pt.up.fe.ni.website.backend.repository.PerActivityRoleRepository
+import pt.up.fe.ni.website.backend.repository.RoleRepository
+
+@Service
+class RoleService(private val roleRepository: RoleRepository, private val perActivityRoleRepository: PerActivityRoleRepository) {
```
BrunoRosendo:
```suggestion
class RoleService(
    private val roleRepository: RoleRepository,
    private val perActivityRoleRepository: PerActivityRoleRepository
) {
```
Same thing about line length

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+
+        if (foundActivity == null) {
+            foundActivity = PerActivityRole(role, activity, Permissions())
+        }
+
+        foundActivity.permissions.add(permission)
+        perActivityRoleRepository.save(foundActivity)
+    }
```
BrunoRosendo:
```suggestion
fun grantPermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
        val foundActivity = activity.perRoles
            .find { it.activity == activity } ?: PerActivityRole(role, activity, Permissions())

        foundActivity.permissions.add(permission)
        perActivityRoleRepository.save(foundActivity)
    }
```

I think this one is definitely worth it

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,58 @@
...
+                foundActivity = perActivity
+                break
+            }
+        }
+
+        if (foundActivity != null) {
+            foundActivity.permissions.remove(permission)
+            perActivityRoleRepository.save(foundActivity)
+        }
+    }
```
BrunoRosendo:
```suggestion
fun revokePermissionFromRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
        val foundActivity = activity.perRoles.find { it.activity == activity } ?: return

        foundActivity.permissions.remove(permission)
        foundActivity.let { perActivityRoleRepository.save(it) }
    }
```
Similarly, you could use the one below:
```kotlin
fun revokePermissionFromRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
        val foundActivity = activity.perRoles.find { it.activity == activity }

        foundActivity?.permissions?.remove(permission)
        foundActivity?.let { perActivityRoleRepository.save(it) }
    }
```

src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
```diff
@@ -56,6 +58,24 @@ class Account(
...
 
+    @JoinColumn
+    @ManyToMany(cascade = [CascadeType.ALL], fetch = FetchType.EAGER)
+    val roles: List<@Valid Role> = emptyList(),
+
     @Id @GeneratedValue
     val id: Long? = null
-)
+) {
+    fun getEffectivePermissionsForActivity(activity: Activity) {
```
BrunoRosendo: Since we have this in the model, it might be a good idea to unit test it. What do you think?

src/test/kotlin/pt/up/fe/ni/website/backend/permissions/PermissionsTest.kt
```diff
@@ -0,0 +1,12 @@
+package pt.up.fe.ni.website.backend.permissions
+
+import org.junit.jupiter.api.Test
+
+class PermissionsTest {
```
BrunoRosendo: We could maybe test `fromLong` and `toLong` here?

src/test/kotlin/pt/up/fe/ni/website/backend/permissions/PermissionsTest.kt
```diff
@@ -0,0 +1,12 @@
...
+
+    @Test
+    fun permissionsShouldBeBuilt() {
```
BrunoRosendo: I didn't quite understand the test, can you explain?

BrunoRosendo (changes requested):
This PR is really f*cking awesome, it will be such an important part of the website!

This is the basis for role management, #42, #60, and more, so let's make sure this is merged well. So I wrote a considerable amount of suggestions, even with the good work so far

src/test/kotlin/pt/up/fe/ni/website/backend/permissions/PermissionsTest.kt
limwa: Nevermind that, I wrote it for debugging things and forgot to remove it!

src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
```diff
@@ -56,6 +58,24 @@ class Account(
     @OneToMany(cascade = [CascadeType.ALL], fetch = FetchType.EAGER)
     val websites: List<@Valid CustomWebsite>,
 
+    @JoinColumn
+    @ManyToMany(cascade = [CascadeType.ALL], fetch = FetchType.EAGER)
```
BrunoRosendo: Should this cascade? I think it means we'll delete the roles when an account is deleted

BrunoRosendo (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
limwa: I don't really like `activityRoles` here. For instance, if I have an `activity` variable, to access the roles, I have to write `activity.activityRoles` which gets kind of repetitive.
What about `activity.associatedRoles`?

BrunoRosendo: You're right, I like associatedRoles

limwa: Awesome!

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -10,6 +10,7 @@ import java.util.Date
 class Event(
     title: String,
     description: String,
+    perRoles: List<PerActivityRole> = emptyList(),
```
limwa: Change this to associatedRoles.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
```diff
@@ -6,6 +6,7 @@ import jakarta.persistence.Entity
 class Project(
     title: String,
     description: String,
+    perRoles: List<PerActivityRole> = emptyList(),
```
limwa: Change this to associatedRoles.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
```diff
@@ -0,0 +1,40 @@
...
+    @JsonProperty(required = true)
+    var isSection: Boolean,
+
+    @JoinColumn
+    @ManyToMany
+    var accounts: List<@Valid Account>,
+
+    @JoinColumn
+    @OneToMany
+    var perActivities: List<@Valid PerActivityRole>,
```
limwa: @BrunoRosendo what about this? Should I leave it as it is?

limwa:
The force push removed the review comments 😫
As far as I remember, the ones left were about lazy/eager fetches and unit tests to Permissions and getEffectivePermissionsForActivity. I'll take care of the tests today and, regarding the fetches, we'll have to discuss it further because I don't really know what I'm doing 😅

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
BrunoRosendo: Associated roles doesn't make sense here, is activityRoles ok?

BrunoRosendo:
> 

The comments are still in the PR page, even if they don't show in the changes

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
BrunoRosendo: Now that I see this again, the last line is weird, perActivityRoleRepository.save(foundActivity) seems better

limwa:
Oh, true, didn't notice that, thanks!

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
limwa: I'll use associatedActivities, since we used associatedRoles on Activity

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
limwa: Thanks!

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
```diff
@@ -21,6 +26,10 @@ abstract class Activity(
     @field:Size(min = Constants.Description.minSize, max = Constants.Description.maxSize)
     open val description: String,
 
+    @JoinColumn
+    @OneToMany(cascade = [CascadeType.ALL], fetch = FetchType.EAGER)
+    open var associatedRoles: List<@Valid PerActivityRole>,
```
BrunoRosendo, line 31: This has nothing to do with this PR but Activities' attributes shouldn't need to be open, can you change it?
Also, isn't val ok here?

src/main/kotlin/pt/up/fe/ni/website/backend/model/permissions/Permissions.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    companion object {
+        fun fromLong(encoded: Long): Permissions {
+            val decodedPermissions = Permission.values()
+                .filter { perm ->
+                    // To decide if the permission is present in the `encoded` Long,
+                    // we need to see if the bits for those permissions are set (equal to 1)
+                    // `encoded ushr n and 1L` returns the value of the `n`-th bit of `encoded`
+                    val encodedBit = encoded ushr perm.bit and 1L
+                    return@filter encodedBit == 1L
```
BrunoRosendo, line 17:
```suggestion
                    encodedBit == 1L
```
This should be the same. Do you prefer the other version because it's explicit?

src/main/kotlin/pt/up/fe/ni/website/backend/model/permissions/Permissions.kt
```diff
@@ -0,0 +1,46 @@
...
+            return Permissions(decodedPermissions)
+        }
+    }
+
+    // To encode the permissions into a Long, we need to set the
+    // corresponding bit to 1, if the permission is present
+    // `1L shl n` returns a number whose binary representation has
+    // a single 1 in the `n`-th position
+    fun toLong() = fold(0L) { acc, perm ->
+        return@fold acc or (1L shl perm.bit)
```
BrunoRosendo: same as above, just asking

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,43 @@
...
+        perActivityRoleRepository.save(foundActivity)
+    }
+
+    fun revokePermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
+        val foundActivity = activity.associatedRoles
+            .find { it.activity == activity } ?: return
+
+        foundActivity.permissions.remove(permission)
+        perActivityRoleRepository.save(foundActivity)
+    }
```
BrunoRosendo: I guess the params would probably be IDs but that can be changed in #79 if you wish

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        foundActivity.permissions.add(permission)
+        perActivityRoleRepository.save(foundActivity)
+    }
+
+    fun revokePermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
+        val foundActivity = activity.associatedRoles
+            .find { it.activity == activity } ?: return
+
+        foundActivity.permissions.remove(permission)
```
BrunoRosendo: You can use removeIf to simplify

BrunoRosendo (changes requested):
Good job with the changes, it's looking great! I only have some details left

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
limwa: I can change it, no worries 
Regarding the use of var, I made it that way since I needed it for the tests and because it follows what we've done on other models (most, if not all, properties are var).

src/main/kotlin/pt/up/fe/ni/website/backend/model/permissions/Permissions.kt
limwa: Yes, I generally prefer the use of return instead of using the version without the return, unless it's a single line operation. I am open to changing that, if you'd like me to.

src/main/kotlin/pt/up/fe/ni/website/backend/model/permissions/Permissions.kt
limwa: Yeah, here I want to change as well, thanks for letting me know!

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
limwa: I can take care of that, no problem!

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
limwa: Thanks!

src/main/kotlin/pt/up/fe/ni/website/backend/model/permissions/Permissions.kt
BrunoRosendo: It's fine this way :)

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
limwa: Actually, it's not the same thing. `removeIf` removes all elements from the collection that fulfill the predicate.
Furthermore, without the return, a null value would still trigger a `save` action on the repository.

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
limwa: Actually, on second thought, it would be best if that was done on the controllers' PR. The use of IDs needs to be evaluated based on the methods required for the controllers.

limwa:
@BrunoRosendo I've applied the change I thought made sense. Let me know what you think about the replies I've written in the review comments.

BrunoRosendo (approved):
Great job!

BrunoRosendo:
Hey guys, this a very important PR so please review 🙏

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    fun grantPermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
+        val foundActivity = activity.associatedRoles
+            .find { it.activity == activity } ?: PerActivityRole(role, activity, Permissions())
+
+        foundActivity.permissions.add(permission)
+        perActivityRoleRepository.save(foundActivity)
+    }
+
+    fun revokePermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
```
ttoino:
```suggestion
    fun revokePermissionFromRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
```

I think this should be renamed for consistency.

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    fun revokePermissionFromRole(role: Role, permission: Permission) {
+        role.permissions.remove(permission)
+        roleRepository.save(role)
+    }
+
+    fun grantPermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
+        val foundActivity = activity.associatedRoles
+            .find { it.activity == activity } ?: PerActivityRole(role, activity, Permissions())
```
ttoino:
```suggestion
        val foundActivity = activity.associatedRoles
            .find { it.role == role } ?: PerActivityRole(role, activity, Permissions())
```

Shouldn't we filter based on the role, as we already know the activity is the requested one?

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        foundActivity.permissions.add(permission)
+        perActivityRoleRepository.save(foundActivity)
+    }
+
+    fun revokePermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
+        val foundActivity = activity.associatedRoles
+            .find { it.activity == activity } ?: return
```
ttoino: Same as above

ttoino (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
```diff
@@ -0,0 +1,40 @@
...
+    @JsonProperty(required = true)
+    var isSection: Boolean,
+
+    @JoinColumn
+    @ManyToMany
+    var accounts: List<@Valid Account> = emptyList(),
+
+    @JoinColumn
+    @OneToMany
+    var associatedActivities: List<@Valid PerActivityRole> = emptyList(),
```
bdmendes, line 35: Why `associatedActivities`? `perActivityRoles` seems more natural to me.

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
bdmendes: I believe so. Perhaps this would be more readable:
```
val foundActivity = roles.perActivityRoles.find { it.activity == activity } ?: PerActivityRole(role, activity, Permissions())
```

bdmendes: This also indicates the need for good testing when the controller is developed.

bdmendes (commented):
Nice work!

ttoino (approved):
LGTM!

ttoino:
You probably need to fix the event tests after #55 was merged.

limwa:
> You probably need to fix the event tests after #55 was merged.

Yep, I think I just took care of that

PR #52, opened 2022-11-16
Added new field "isArchived" on Project class

### LINKED ISSUE #40: projects: add support for archived projects
We should have a boolean field (e.g. `isArchived`) informing if the project is archived or not.

This field should default to false when a project is created and be changed by a PUT endpoint.

coutinho21:
Closes #40 

Added a new field on the Project and ProjectDto class "isArchived" to identify if a project is archived or not. Also updated the update project test with that new field. 

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -10,7 +10,6 @@ import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import pt.up.fe.ni.website.backend.model.dto.ProjectDto
 import pt.up.fe.ni.website.backend.service.ProjectService
-import java.util.spi.LocaleNameProvider
 
 @RestController
 @RequestMapping("/projects")
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -33,9 +33,9 @@ class ProjectController(private val service: ProjectService) {
         @RequestBody dto: ProjectDto
     ) = service.updateProjectById(id, dto)
 
-    @PutMapping("/archive/{id}")
+    @PutMapping("/{id}/archive")
     fun archiveProjectById(@PathVariable id: Long) = service.archiveProjectById(id)
 
-    @PutMapping("/unarchive/{id}")
+    @PutMapping("/{id}/unarchive")
     fun unarchiveProjectById(@PathVariable id: Long) = service.unarchiveProjectById(id)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -10,6 +10,7 @@ import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import pt.up.fe.ni.website.backend.model.dto.ProjectDto
 import pt.up.fe.ni.website.backend.service.ProjectService
+import java.util.spi.LocaleNameProvider
 
 @RestController
 @RequestMapping("/projects")
@@ -32,4 +33,10 @@ class ProjectController(private val service: ProjectService) {
         @PathVariable id: Long,
         @RequestBody dto: ProjectDto
     ) = service.updateProjectById(id, dto)
+
+    @PutMapping("/archive/{id}")
+    fun archiveProjectById(@PathVariable id: Long) = service.archiveProjectById(id)
+
+    @PutMapping("/unarchive/{id}")
+    fun unarchiveProjectById(@PathVariable id: Long) = service.unarchiveProjectById(id)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
@@ -17,6 +17,9 @@ class Project(
     @field:Size(min = Constants.Description.minSize, max = Constants.Description.maxSize)
     var description: String,
 
+    @JsonProperty(required = false)
+    var isArchived: Boolean = false,
+
     @Id @GeneratedValue
     val id: Long? = null
 )
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
@@ -17,7 +17,6 @@ class Project(
     @field:Size(min = Constants.Description.minSize, max = Constants.Description.maxSize)
     var description: String,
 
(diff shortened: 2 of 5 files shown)


src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/ProjectDto.kt
```diff
@@ -4,5 +4,6 @@ import pt.up.fe.ni.website.backend.model.Project
 
 class ProjectDto(
     val name: String,
-    val description: String
+    val description: String,
+    val isArchived: Boolean = false
```
bdmendes, line 8: Keep in mind that this makes sense if it's ok to assume that if a project is being modified, then it's not supposed to be archived. Say, if a user updates the name of an archived project and does not pass `isArchived`, it will default to false and the project will be unarchived. If this is intended behavior, it's ok for me

bdmendes (approved):

BrunoRosendo (changes requested):
Good job so far! I think this implementation works as intended

However, this issue also asks you to implement some endpoints to archive and unarchive projects. Can you please look at that? Some tests for those endpoints would be nice as well

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/ProjectDto.kt
BrunoRosendo: Good point. If, for some reason, it's not given in the edit endpoint, then I think the `isArchived` field should stay as it was. Should we add some logic to the edit endpoint or is there a more elegant way?

I'm happy to hear your thoughts on this @bdmendes @coutinho21

src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
```diff
@@ -17,6 +17,9 @@ class Project(
     @field:Size(min = Constants.Description.minSize, max = Constants.Description.maxSize)
     var description: String,
 
+    @JsonProperty(required = false)
+    var isArchived: Boolean = false,
```
BrunoRosendo: Two questions:
- Why is this not required? The user doesn't need to send it but we always have it when converting from our Dto
- Is there a reason why we have the default value here as well as in the Dto? Is the Dto not enough?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/ProjectControllerTest.kt
```diff
@@ -296,4 +300,58 @@ internal class ProjectControllerTest @Autowired constructor(
...
+    @Nested
+    @DisplayName("PUT /projects/unarchive/{projectId}")
+    inner class UnarchiveProject {
+        @BeforeEach
+        fun addProject() {
+            repository.save(testProject)
+        }
+
+        @Test
+        fun `should archive the project`() {
```
BrunoRosendo:
```suggestion
        fun `should unarchive the project`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/ProjectControllerTest.kt
```diff
@@ -296,4 +300,58 @@ internal class ProjectControllerTest @Autowired constructor(
...
+        }
+    }
+
+    @Nested
+    @DisplayName("PUT /projects/unarchive/{projectId}")
+    inner class UnarchiveProject {
+        @BeforeEach
+        fun addProject() {
+            repository.save(testProject)
+        }
```
BrunoRosendo: I think we should create an archived project here. So we can properly test the endpoint

BrunoRosendo (changes requested):
Good job with the endpoints and the tests! This is almost ready to merge, just a couple of details

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -32,4 +32,10 @@ class ProjectController(private val service: ProjectService) {
         @PathVariable id: Long,
         @RequestBody dto: ProjectDto
     ) = service.updateProjectById(id, dto)
+
+    @PutMapping("/archive/{id}")
+    fun archiveProjectById(@PathVariable id: Long) = service.archiveProjectById(id)
+
+    @PutMapping("/unarchive/{id}")
+    fun unarchiveProjectById(@PathVariable id: Long) = service.unarchiveProjectById(id)
```
BrunoRosendo:
```suggestion
    @PutMapping("/{id}/archive")
    fun archiveProjectById(@PathVariable id: Long) = service.archiveProjectById(id)

    @PutMapping("/{id}/unarchive")
    fun unarchiveProjectById(@PathVariable id: Long) = service.unarchiveProjectById(id)
```
This is a bit picky but the operation usually comes at the end (it's apparently considered more RESTful). You can look at examples of public APIs:
https://docs.github.com/en/rest/gists/gists#star-a-gist
https://developer.twitter.com/en/docs/twitter-api/tweets/likes/migrate

BrunoRosendo (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
coutinho21: You're right, the default value in the Dto is enough.

coutinho21: Actually, when i remove the default value in the project class, the tests asks me to give one. 

image

coutinho21: I decided to put the default value in the Project class and remove it from the Dto. Waiting for feedback.

limwa: Have you tested using `val isArchived: Boolean? = false,` in ProjectDto?
I'm suggesting this because I think it makes sense to mark isArchived as required, however, I don't know how that solution would behave if a null value were to be passed on isArchived.

limwa: Another option would be to leave it as it is and use @JsonProperty(required = false) on the ProjectDto just so it is clear to anybody looking at the code that it isn't a required argument and to put @JsonProperty(required = true) on the project itself

coutinho21: Why would isArchived be required on the project itself? I think it should not be required. When a project is created, it should not be archived by default.

BrunoRosendo: The change in ProjectDto actually doesn't change any behaviour. `Boolean` is treated as a Java primitive and it always has a default value of `false` (even if you don't write it explicitly). I would personally leave the `= false` for readability.

As for the second discussion, @coutinho21 and @limwa , the `JsonProperty(required = true)` annotation doesn't mean that the `isArchived` field should be true by default. It means that, when converting a JSON object to a `Project`, the `isArchived` field does not need to be present. Since we have a default value, I believe it should NOT be required. However, since that's the default behaviour, you don't even need to put the annotation

BrunoRosendo (approved):
Good job!

PR #54, opened 2022-11-17
Feature/auth

### LINKED ISSUE #12: accounts: implement authentication
Based on defined roles, authentication is mandatory for the backoffice and possibly to limit views, too.

bdmendes:
Closes #12

# Review checklist


bdmendes:
The stateless implementation is working. We still need to save state, either to invalidate tokens or to link tokens to users for a manual check of permissions. How would you approach this @limwa?

bdmendes:
After some research, I found that JWT with total stateless behavior is the best way to go (see https://stackoverflow.com/questions/6068113/do-sessions-really-violate-restfulness).
This implementation uses access and refresh tokens. An access token contains the user roles and is to be used as a bearer token in the Authorization header. It has a short (configurable) expiration time of 60 minutes, so that invalid roles aren't allowed for much time after they get removed. For improving the user's UX on the frontend (and not force the frontend to save the password), it will save a refresh token that will be valid for 7 days and will be used to generate a new access token (with updated rules and more 60 minutes) every time it needs to. This refresh stuff should really be implemented in some Vue/React/wtv lib, although I did not do much research on this. In the worst case scenario, we can simply create our own thin wrapper around requests logic.

bdmendes:
I'll soon add a custom message for invalid tokens. Meanwhile, everything that is important is done, so this is ready for review

bdmendes:
This is fully ready

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.security.oauth2.server.resource.authentication.JwtGrantedAuthoritiesConverter
+import org.springframework.security.web.SecurityFilterChain
+import org.springframework.web.cors.CorsConfiguration
+import org.springframework.web.cors.UrlBasedCorsConfigurationSource
+import org.springframework.web.filter.CorsFilter
+import org.springframework.web.servlet.HandlerExceptionResolver
+
+@Configuration
+@EnableWebSecurity
+@EnableGlobalMethodSecurity(prePostEnabled = true)
```
BrunoRosendo, line 29: I didn't quite understand this annotation. Does `preprePostEnabled` allow us to use annotations such as `PreAuthorize`? Also, what's the difference between this option and jsr250?

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
```diff
@@ -0,0 +1,82 @@
...
+class AuthConfig(
+    val authConfigProperties: AuthConfigProperties,
+    @Qualifier("handlerExceptionResolver") val exceptionResolver: HandlerExceptionResolver
```
BrunoRosendo, line 32: Where does `handlerExceptionResolver` come from? I don't see it specified anywhere

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
```diff
@@ -0,0 +1,82 @@
...
+) {
+    @Bean
+    fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
+        return http.csrf { csrf -> csrf.disable() }.cors().and()
```
BrunoRosendo, line 36: We have #58 for cors but should we disable csrf?

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
```diff
@@ -0,0 +1,82 @@
...
+        config.addAllowedOrigin("*")
+        config.addAllowedHeader("*")
+        config.addAllowedMethod("*")
+        source.registerCorsConfiguration("/**", config)
+        return CorsFilter(source)
+    }
+
+    fun rolesConverter(): JwtAuthenticationConverter? {
+        val authoritiesConverter = JwtGrantedAuthoritiesConverter()
+        authoritiesConverter.setAuthorityPrefix("ROLE_")
```
BrunoRosendo, line 77: Does this mean all roles will start with `ROLE_`?

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
```diff
@@ -0,0 +1,82 @@
...
+            .and().authenticationEntryPoint { request, response, exception ->
+                exceptionResolver.resolveException(request, response, null, exception)
+            }.and()
+            .sessionManagement { session -> session.sessionCreationPolicy(SessionCreationPolicy.STATELESS) }
+            .httpBasic().disable().build()
+    }
+
+    @Bean
+    fun jwtDecoder(): JwtDecoder {
+        return NimbusJwtDecoder.withPublicKey(authConfigProperties::publicKey.get()).build()
```
BrunoRosendo, line 48: What's the difference between `authConfigProperties::publicKey` and `authConfigProperties.publicKey`?

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfigProperties.kt
```diff
@@ -0,0 +1,15 @@
...
+import java.security.interfaces.RSAPublicKey
+
+@ConstructorBinding
+@ConfigurationProperties(prefix = "auth")
+data class AuthConfigProperties(
+    val publicKey: RSAPublicKey,
+    val privateKey: RSAPrivateKey,
+    val jwtAccessExpirationMinutes: Long,
+    val jwtRefreshExpirationDays: Long
+)
```
BrunoRosendo, line 15: I didn't know about this, quite awesome!

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -0,0 +1,44 @@
...
+import javax.servlet.http.HttpServletRequest
+
+data class LoginDto(
+    val email: String,
+    val password: String
+)
+
+data class TokenDto(
+    val token: String
+)
```
BrunoRosendo, line 18: We should think about how we want to store these DTOs. Until now, we only used DTOs that directly match our models so they're in the `model/` package. However, this is not the case here.

I don't have a solution right now so I'll leave some food for thought. Should we keep all DTOs/DAOs in their own package and rename our current `Dto` abstract class to `ModelDto` or something of sorts? Or should we keep our current implementation and just move these new type of Dtos to their own package?

Please let em know what you think

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
```diff
@@ -91,7 +95,25 @@ class ErrorController : ErrorController {
...
+    @ExceptionHandler(AccessDeniedException::class)
+    @ResponseStatus(HttpStatus.FORBIDDEN)
+    fun forbidden(e: AccessDeniedException): CustomError {
+        return wrapSimpleError(e.message ?: "you don't have permission to access this resource")
+    }
+
+    @ExceptionHandler(AuthenticationException::class)
+    @ResponseStatus(HttpStatus.UNAUTHORIZED)
+    fun invalidAuthentication(e: AuthenticationException): CustomError {
+        return wrapSimpleError(e.message ?: "invalid authentication")
```
BrunoRosendo, line 108: It's a detail but I don't think `invalid authentication` correctly represents this exception. A user probably didn't provide any credentials, so there was simply none instead of invalid authentication. 401 specified that it lacks valid authentication credentials (see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401)), not that it has invalid ones.

Maybe change it to simply "unauthorized"?

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/AccountDto.kt
```diff
@@ -4,12 +4,13 @@ import pt.up.fe.ni.website.backend.model.Account
...
     val email: String,
+    val password: String,
+    val name: String,
     val bio: String?,
     val birthDate: Date?,
     val photoPath: String?,
     val linkedin: String?,
     val github: String?,
-    val websites: List<CustomWebsiteDto>
+    val websites: List<CustomWebsiteDto>?
```
BrunoRosendo, line 15: Why is this nullable now? Is it ok to pass a null value when it's required in the account model?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -0,0 +1,89 @@
...
+    val accountService: AccountService,
+    val authConfigProperties: AuthConfigProperties,
+    val jwtEncoder: JwtEncoder,
+    val jwtDecoder: JwtDecoder,
+    private val passwordEncoder: PasswordEncoder
+) {
+    fun authenticate(email: String, password: String): Account {
+        val account = accountService.getAccountByEmail(email)
+        if (!passwordEncoder.matches(password, account.password)) {
+            throw ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY, "invalid credentials")
```
BrunoRosendo: Once again, a bit picky comment. I feel uneasy specifying a static code in the services. I think the whole point of throwing exceptions in the services and handling them in `ErrorController` is to separate business logic from the responses sent to the client and to make sure the same exception is tied to a single status code.

This is my personal view of the issue but there's some discussion around it (https://www.baeldung.com/exception-handling-for-rest-with-spring). An advantage of using ResponseStatusException is, obviously, reducing code size. I will leave the decision to you but I wanted to point this out

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -0,0 +1,89 @@
...
+        return generateToken(account, Duration.ofDays(authConfigProperties.jwtRefreshExpirationDays), true)
+    }
+
+    fun refreshAccessToken(refreshToken: String): String {
+        val jwt =
+            try {
+                jwtDecoder.decode(refreshToken)
+            } catch (e: Exception) {
+                throw ResponseStatusException(HttpStatus.UNAUTHORIZED, "invalid refresh token")
+            }
```
BrunoRosendo: Continuing the comment above, global exception handling also simplifies services in cases like this
```suggestion
        val jwt =  jwtDecoder.decode(refreshToken)
```

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    @PostMapping("/refresh")
+    fun refreshAccessToken(@RequestBody tokenDto: TokenDto): Map<String, String> {
+        val accessToken = authService.refreshAccessToken(tokenDto.token)
+        return mapOf("access_token" to accessToken)
+    }
+
+    @GetMapping
+    @PreAuthorize("hasRole('MEMBER')")
+    fun checkAuthentication(request: HttpServletRequest): Map<String, String> {
```
BrunoRosendo: Why is `request` needed when it's not being used inside the function? (I think it is used somewhere, since my IDE tells me so, I just don't know how)

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -0,0 +1,89 @@
...
+            .map(GrantedAuthority::getAuthority)
+            .collect(Collectors.joining(" "))
+        val currentInstant = Instant.now()
+        val claims = JwtClaimsSet
+            .builder()
+            .issuer("self")
+            .issuedAt(currentInstant)
+            .expiresAt(currentInstant.plus(expiration))
+            .subject(account.email)
+            .claim("scope", scope)
```
BrunoRosendo, line 78: If I understood well, the claim is composed of a scope which shows the user's roles?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -373,4 +396,22 @@ class AccountControllerTest @Autowired constructor(
...
+                "name" to this?.name,
+                "email" to this?.email,
+                "password" to this?.password,
+                "bio" to this?.bio,
+                "birthDate" to this?.birthDate.toJson(),
+                "photoPath" to this?.photoPath,
+                "linkedin" to this?.linkedin,
+                "github" to this?.github,
+                "websites" to this?.websites
+            )
```
```suggestion
            objectMapper.convertValue(this, Map::class.java).plus(
                mapOf(
                    "password" to this?.password
                )
            )
        )
```
This seems easier to maintain

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -0,0 +1,89 @@
...
+            .build()
+        return jwtEncoder.encode(JwtEncoderParameters.from(claims)).tokenValue
+    }
+
+    private fun getAuthorities(account: Account): List<GrantedAuthority> {
+        return listOf("BOARD", "MEMBER").stream() // TODO: get roles from account
```
BrunoRosendo, line 84: We should create an issue for this

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
```diff
@@ -0,0 +1,185 @@
...
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+class AuthControllerTest @Autowired constructor(
+    val repository: AccountRepository,
+    val mockMvc: MockMvc,
+    val objectMapper: ObjectMapper,
+    passwordEncoder: PasswordEncoder
+) {
+    final val testPassword = "testPassword"
+
+    // TODO: Make sure to add "MEMBER" role to the account
```
BrunoRosendo, line 38: This can be referenced in the issue mentioned above

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
```diff
@@ -0,0 +1,185 @@
...
+    @DisplayName("POST /auth/new")
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    inner class GetNewToken {
+        @BeforeAll
+        fun setup() {
+            repository.save(testAccount)
+        }
+
+        @Test
+        fun `should fail when email is invalid`() {
```
BrunoRosendo:
```suggestion
        fun `should fail when email is not registered`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
```diff
@@ -0,0 +1,185 @@
...
+                status { isUnprocessableEntity() }
+                jsonPath("$.errors[0].message") { value("invalid credentials") }
+            }
+        }
+
+        @Test
+        fun `should return access and refresh tokens`() {
+            mockMvc.post("/auth/new") {
+                contentType = MediaType.APPLICATION_JSON
+                content = objectMapper.writeValueAsString(LoginDto(testAccount.email, testPassword))
```
BrunoRosendo, line 93: We should use the same method here and in the test above. I personally like the first one more because it doesn't depend on the Dto implementation but it can also be more verbose

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
```diff
@@ -0,0 +1,185 @@
...
+                mockMvc.get("/auth") {
+                    header("Authorization", "Bearer $accessToken")
+                }.andExpect {
+                    status { isOk() }
+                    jsonPath("$.authenticated_user") { value(testAccount.email) }
+                }
+            }
+        }
+
+        // TODO: Add tests for role access when implemented
```
BrunoRosendo: I don't think this TODO is necessary, tests should always be implemented in a new feature

src/test/kotlin/pt/up/fe/ni/website/backend/utils/ValidationTester.kt
```diff
@@ -1,4 +1,4 @@
-package pt.up.fe.ni.website.backend.controller
+package pt.up.fe.ni.website.backend.utils
```
BrunoRosendo, line 1: nice catch

BrunoRosendo (changes requested):
This is a really great job! Amazing PR, this will be extremely useful going forward :rocket: 

That said, I wrote quite a few comments, mostly asking questions since this is a complex PR. I also left some suggestions regarding smaller details.

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
bdmendes: `prepostEnabled` enables the most powerful `PreAuthorize` annotation. Check this out: https://stackoverflow.com/questions/43961625/rolesallowed-vs-preauthorize-vs-secured

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
bdmendes: Two beans are registered for the `HandlerExceptionResolver` type, so it needs to be disambiguated this way. We need to dispatch the exception manually to this resolver, otherwise, our `ErrorController` won't catch this. Check this out: https://stackoverflow.com/questions/19767267/handle-spring-security-authentication-exceptions-with-exceptionhandler

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
bdmendes: No. This overrides the default oauth2 behavior that prepends `SCOPE_` to all authorities. We need the `ROLE_` prefix prepended in the internal authorities to use `PreAuthorize(hasRole(x))`. x is stored internally with ROLE_x.

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
bdmendes: I don't think there is any difference

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
bdmendes: I do like this:
```
dto/
     model/
     auth/
     otherSemanticUseCase/
```
Should I do this here?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
bdmendes: This actually gets triggered only when a bearer token is sent and it is unable to be parsed (is actually invalid). If no auth is provided, Spring Security raises `AccessDeniedException`, which is handled by `forbidden`, directly on top of this. So I think this is a fine message.

src/main/kotlin/pt/up/fe/ni/website/backend/model/dto/AccountDto.kt
bdmendes: Yes, the default list constructor is used instead upon the model creation.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
bdmendes: I agree. I'll throw a more semantic exception in the next commit.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
bdmendes: I'll keep the invalid/expired logic and throw a more semantic exception.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
bdmendes: Thanks, it got forgotten after a refactor

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
bdmendes: Yes, in this implementation yes. There is some discussion against it: https://stackoverflow.com/questions/47224931/is-setting-roles-in-jwt-a-best-practice. However, in our use case, to reduce database fetches and code complexity, I think it is a worthy, small security tradeoff.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
bdmendes: Awesome!

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
bdmendes: Thanks

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
bdmendes: Right!

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
bdmendes: I'll convert the first to this one since I prefer the less verbose approach

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
bdmendes: See #60

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
bdmendes: Sure

src/main/kotlin/pt/up/fe/ni/website/backend/config/auth/AuthConfig.kt
bdmendes: Agreed, although that complicates our requests quite a lot (see https://www.baeldung.com/spring-security-csrf). I have created #61 for that.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
BrunoRosendo: I like that as well. I think we should do it here, so we don't merge code that's going to change immediately

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
BrunoRosendo: I see, I think it makes sense then. Thanks for explaining

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -49,10 +48,10 @@ class AuthService(
             try {
                 jwtDecoder.decode(refreshToken)
             } catch (e: Exception) {
-                throw ResponseStatusException(HttpStatus.UNAUTHORIZED, "invalid refresh token")
+                throw InvalidBearerTokenException("invalid refresh token")
             }
         if (jwt.expiresAt?.isBefore(Instant.now()) != false) {
-            throw ResponseStatusException(HttpStatus.UNAUTHORIZED, "refresh token has expired")
+            throw InvalidBearerTokenException("refresh token has expired")
```
BrunoRosendo, line 54: I didn't test this on my own so I'm going to ask. I like the change but is this exception being handled or does it just return an unexpected error?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
BrunoRosendo: I see, I don't think it should be a big concern for us

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
bdmendes: Yes, it is an AuthenticationException. This is being tested, too

BrunoRosendo (approved):
Great job! :)

PR #55, opened 2022-11-23
Add more properties and endpoints to events

### LINKED ISSUE #5: events: add more information to model and more endpoints
Currently, the event model is very basic and doesn't contain all the information we'll need. The first part of the issue is to think about this and come up with new fields that the Event Model should have.

Then, you can proceed to creating new endpoints in the Event Controller. These could include things like retrieving information on a single event, editing an event and deleting one, amongst others you may find useful.

ttoino:
Closes #5 

Added a few new properties to the event model and dto, namely when it happens (startDate and endDate), where it happens (location), and its category and thumbnailPath.

Also added endpoints for getting events by id and by category, and to update and delete an event.

# Review checklist


src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -19,7 +22,24 @@ class Event(
...
+
+    @field:NullOrNotBlank
+    @field:URL
+    val url: String?,
+
+    @field:Size(min = Constants.Location.minSize, max = Constants.Location.maxSize)
+    val location: String?,
+
+    @field:Size(min = Constants.Category.minSize, max = Constants.Category.maxSize)
+    val category: String?,
```
BrunoRosendo, line 31: I'm not against it but what is your idea behind the category field? What values would it have on the website?
I'm asking so we don't merge unnecessary code

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -66,9 +81,103 @@ internal class EventControllerTest @Autowired constructor(
...
+            for (event in testEvents) repository.save(event)
+        }
+
+        @Test
+        fun `should return all events of the category`() {
+            mockMvc.get("/events/category/${testEvent.category}")
+                .andExpect {
+                    status { isOk() }
+                    content { contentType(MediaType.APPLICATION_JSON) }
+                    jsonPath("$.length()") { value(1) }
```
BrunoRosendo: This is already well but I'll just add a minor suggestion. Maybe it'd be better to make a test that returns 2 events, so that we're sure the system is really returning all events of that category (and not just one)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -66,9 +81,103 @@ internal class EventControllerTest @Autowired constructor(
...
+        }
+    }
+
     @Nested
     @DisplayName("POST /events/new")
     inner class CreateEvent {
+        @AfterEach
+        fun clearEvents() {
+            repository.deleteAll()
+        }
```
BrunoRosendo: Why is this now needed? Creating posts should not depend on existing ones

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -19,7 +22,24 @@ class Event(
     val description: String,
 
     @JsonProperty(required = true)
-    val date: Date,
+    val startDate: Date,
+
+    val endDate: Date?,
```
BrunoRosendo: What do you think about adding a validation to check if endDate is after startDate?
We could perhaps create a new annotation for this (annotations/validation).

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -151,6 +270,340 @@ internal class EventControllerTest @Autowired constructor(
...
+            @DisplayName("url")
+            @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+            inner class UrlValidation {
+                @BeforeAll
+                fun setParam() {
+                    validationTester.param = "url"
+                }
+
+                @Test
+                fun `should be a URL`() = validationTester.isUrl()
```
BrunoRosendo: I think this is missing the `validationTester.isNullOrNotBlank()` test

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -151,6 +270,340 @@ internal class EventControllerTest @Autowired constructor(
...
+        }
+    }
+
+    @Nested
+    @DisplayName("DELETE /events/{eventId}")
+    inner class DeleteEvent {
+        @BeforeEach
+        fun addEvent() {
+            repository.save(testEvent)
+        }
```
BrunoRosendo: Note that this will create an event for every single test we make in the future. We could have some logic checking if it already exists but this should work fine in this endpoint

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -151,6 +270,340 @@ internal class EventControllerTest @Autowired constructor(
...
+            @DisplayName("url")
+            @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+            inner class UrlValidation {
+                @BeforeAll
+                fun setParam() {
+                    validationTester.param = "url"
+                }
+
+                @Test
+                fun `should be a URL`() = validationTester.isUrl()
```
BrunoRosendo: Same thing as above

BrunoRosendo (changes requested):
Good job with this PR! I left some small comments

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
ttoino: That makes sense, I've added another event with that category.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
ttoino: You're right, it isn't needed. But I'm pretty sure I copied that from another test file <.<

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
ttoino: You're right! I've added it.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
ttoino: I've left as it is for now, as it's how we're doing it for the other tests.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
ttoino: I based it on the website mock, where the events are separated by category (SINF, Workshops, Jantares de Curso). I think it might make sense to separate the category into a separate model tho, so we can maybe reuse it to categorize projects or other models where it makes sense to do so, and so we're not susceptible to typos when changing an event's category.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
ttoino: I've now implemented a class annotation that achieves that. I'm not really happy with the error it produces (it uses the whole model as the value), but I don't really know how I could change that.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
BrunoRosendo: I get it, it makes sense. Do you think it's worth doing that right now? We could maybe create an issue for later

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
BrunoRosendo: I see... I understand your approach of targeting the class because we're validating two fields. However, this is out of scope of validations and that's why it doesn't seem possible to override the behaviour. A validation error should always be linked with some field (can you imagine handling special cases in the frontend?), this way we have 2 cases:
- The `errors` field is an array of validations errors that can be displayed in some kind of form
- The `error` field is a message referring to the whole request, which can be displayed in a snackbar for example.

In this case, I think we should simplify our lives and do what would be more convenient for the frontend. Imo, this still makes sense as a validation error displayed under the `endDate` field. Therefore, I would change the annotation to target a field and use the `endDate` as the value and param

ttoino: I've been searching and I think it's impossible to access the value of the startDate field inside another field's annotation.

I have found a way to change the propertyPath field of the generated violation, but changing the value field seems to be impossible.

image

I think this is more acceptable but still not perfect :/

BrunoRosendo: With the suggestion of @limwa, I played a bit with `@Embeddable` classes, and I managed to achieve this error:
```
{
    "errors": [
        {
            "message": "endDate must be after startDate",
            "param": "dateInterval",
            "value": {
                "startDate": "28-07-2022",
                "endDate": "27-07-2022"
            }
        }
    ]
}
```

I think this is what we want. What do you think?

ttoino: I think that looks much better, it's now implemented. I also tried playing with `@JsonUnwrapped` to get rid of the nesting, but that would mean that the `param` field would be wrong, so I decided against it.

BrunoRosendo: I think either way is fine. I don't dislike `@JsonUnwrapped`, we could even think about changing the `param` to `endDate`, maybe it'd help the frontend side of things since the error will be displayed in a form. I'll let you choose

src/main/kotlin/pt/up/fe/ni/website/backend/annotations/validation/DateInterval.kt
```diff
@@ -0,0 +1,39 @@
...
+import jakarta.validation.Payload
+import java.util.Date
+import kotlin.reflect.KClass
+import kotlin.reflect.full.memberProperties
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+@Constraint(validatedBy = [DateIntervalValidator::class])
+@MustBeDocumented
+annotation class DateInterval(
```
BrunoRosendo: Perhaps change the annotation's name to avoid confusion with the Embeddable class. Something like `ValidDateInterval` maybe?

src/test/kotlin/pt/up/fe/ni/website/backend/utils/ValidationTester.kt
```diff
@@ -126,6 +126,27 @@ class ValidationTester(
             }
     }
 
+    fun isDateInterval() {
```
BrunoRosendo:
```suggestion
    fun isValidDateInterval() {
```
Same as above

BrunoRosendo (changes requested):
small cosmetic comment

BrunoRosendo:
@ttoino Is this ready? And are the tests failing because of the problem fixed in #80 ? If so, please review so we can merge it asap

ttoino:
> @ttoino Is this ready? And are the tests failing because of the problem fixed in #80 ? If so, please review so we can merge it asap

I think it's ready, except for the failing tests. I tried applying the changes in #80 and the tests are still failing :/

BrunoRosendo:
> > @ttoino Is this ready? And are the tests failing because of the problem fixed in #80 ? If so, please review so we can merge it asap
> 
> I think it's ready, except for the failing tests. I tried applying the changes in #80 and the tests are still failing :/

I managed to get them working by changing `ConstraintValidator<ValidDateInterval, Any>` to `ConstraintValidator<ValidDateInterval, DateInterval>`. I don't understand why it was a problem but I think this is more simpler anyway. This is how it looks in the end

```Kotlin
@Target(AnnotationTarget.CLASS)
@Retention(AnnotationRetention.RUNTIME)
@Constraint(validatedBy = [DateIntervalValidator::class])
@MustBeDocumented
annotation class ValidDateInterval(
    val message: String = "{date_interval.error}",
    val groups: Array<KClass<*>> = [],
    val payload: Array<KClass<Payload>> = []
)

class DateIntervalValidator : ConstraintValidator<ValidDateInterval, DateInterval> {
    override fun isValid(value: DateInterval, context: ConstraintValidatorContext?): Boolean {
        return value.endDate == null || value.endDate.after(value.startDate)
    }
}
```

limwa:
> > > @ttoino Is this ready? And are the tests failing because of the problem fixed in #80 ? If so, please review so we can merge it asap
> > 
> > I think it's ready, except for the failing tests. I tried applying the changes in #80 and the tests are still failing :/
> 
> I managed to get them working by changing `ConstraintValidator<ValidDateInterval, Any>` to `ConstraintValidator<ValidDateInterval, DateInterval>`. I don't understand why it was a problem but I think this is more simpler anyway. This is how it looks in the end
> 
> ```Kotlin
> @Target(AnnotationTarget.CLASS)
> @Retention(AnnotationRetention.RUNTIME)
> @Constraint(validatedBy = [DateIntervalValidator::class])
> @MustBeDocumented
> annotation class ValidDateInterval(
>     val message: String = "{date_interval.error}",
>     val groups: Array<KClass<*>> = [],
>     val payload: Array<KClass<Payload>> = []
> )
> 
> class DateIntervalValidator : ConstraintValidator<ValidDateInterval, DateInterval> {
>     override fun isValid(value: DateInterval, context: ConstraintValidatorContext?): Boolean {
>         return value.endDate == null || value.endDate.after(value.startDate)
>     }
> }
> ```

@BrunoRosendo have you tried `out Any` instead of `Any`? Reading your comment, I realized that might be the cause, but I don't know if Spring supports this

BrunoRosendo:
> > > > @ttoino Is this ready? And are the tests failing because of the problem fixed in #80 ? If so, please review so we can merge it asap
> > > 
> > > 
> > > I think it's ready, except for the failing tests. I tried applying the changes in #80 and the tests are still failing :/
> > 
> > 
> > I managed to get them working by changing `ConstraintValidator<ValidDateInterval, Any>` to `ConstraintValidator<ValidDateInterval, DateInterval>`. I don't understand why it was a problem but I think this is more simpler anyway. This is how it looks in the end
> > ```kotlin
> > @Target(AnnotationTarget.CLASS)
> > @Retention(AnnotationRetention.RUNTIME)
> > @Constraint(validatedBy = [DateIntervalValidator::class])
> > @MustBeDocumented
> > annotation class ValidDateInterval(
> >     val message: String = "{date_interval.error}",
> >     val groups: Array<KClass<*>> = [],
> >     val payload: Array<KClass<Payload>> = []
> > )
> > 
> > class DateIntervalValidator : ConstraintValidator<ValidDateInterval, DateInterval> {
> >     override fun isValid(value: DateInterval, context: ConstraintValidatorContext?): Boolean {
> >         return value.endDate == null || value.endDate.after(value.startDate)
> >     }
> > }
> > ```
> 
> @BrunoRosendo have you tried `out Any` instead of `Any`? Reading your comment, I realized that might be the cause, but I don't know if Spring supports this

I didn't. Isn't `out` just a restriction when declaring generics? I don't it's possible when defining the child class (`Projections are not allowed for immediate arguments of a supertype`)

limwa:
out isn't a restriction per say. List<String> can't be assigned to List<Any> but it can be assigned to List<out Any> (List of things that extend Any). I thought this might have had something to do with why it wasn't working but I guess not.

ttoino:
> 



> > > @ttoino Is this ready? And are the tests failing because of the problem fixed in #80 ? If so, please review so we can merge it asap
> > 
> > 
> > I think it's ready, except for the failing tests. I tried applying the changes in #80 and the tests are still failing :/
> 
> I managed to get them working by changing `ConstraintValidator<ValidDateInterval, Any>` to `ConstraintValidator<ValidDateInterval, DateInterval>`. I don't understand why it was a problem but I think this is more simpler anyway. This is how it looks in the end
> 
> ```kotlin
> @Target(AnnotationTarget.CLASS)
> @Retention(AnnotationRetention.RUNTIME)
> @Constraint(validatedBy = [DateIntervalValidator::class])
> @MustBeDocumented
> annotation class ValidDateInterval(
>     val message: String = "{date_interval.error}",
>     val groups: Array<KClass<*>> = [],
>     val payload: Array<KClass<Payload>> = []
> )
> 
> class DateIntervalValidator : ConstraintValidator<ValidDateInterval, DateInterval> {
>     override fun isValid(value: DateInterval, context: ConstraintValidatorContext?): Boolean {
>         return value.endDate == null || value.endDate.after(value.startDate)
>     }
> }
> ```

I agree it's simpler, but I'm still getting the same error with that code...

BrunoRosendo:
> > 
> 
> > > > @ttoino Is this ready? And are the tests failing because of the problem fixed in #80 ? If so, please review so we can merge it asap
> > > 
> > > 
> > > I think it's ready, except for the failing tests. I tried applying the changes in #80 and the tests are still failing :/
> > 
> > 
> > I managed to get them working by changing `ConstraintValidator<ValidDateInterval, Any>` to `ConstraintValidator<ValidDateInterval, DateInterval>`. I don't understand why it was a problem but I think this is more simpler anyway. This is how it looks in the end
> > ```kotlin
> > @Target(AnnotationTarget.CLASS)
> > @Retention(AnnotationRetention.RUNTIME)
> > @Constraint(validatedBy = [DateIntervalValidator::class])
> > @MustBeDocumented
> > annotation class ValidDateInterval(
> >     val message: String = "{date_interval.error}",
> >     val groups: Array<KClass<*>> = [],
> >     val payload: Array<KClass<Payload>> = []
> > )
> > 
> > class DateIntervalValidator : ConstraintValidator<ValidDateInterval, DateInterval> {
> >     override fun isValid(value: DateInterval, context: ConstraintValidatorContext?): Boolean {
> >         return value.endDate == null || value.endDate.after(value.startDate)
> >     }
> > }
> > ```
> 
> I agree it's simpler, but I'm still getting the same error with that code...

What's the error?

BrunoRosendo:
@ttoino Any progress on this?

ttoino:
> @ttoino Any progress on this?

I'm still getting the same error referring to the DateIntervalValidator. Debugging the code it seems that Spring is having trouble instantiating a "metaDataSourceAdvisor" bean somewhere deep in the implementation, but it doesn't seem to have this issue when instantiating a NullOrNotBlankValidator.

BrunoRosendo:
> > @ttoino Any progress on this?
> 
> I'm still getting the same error referring to the DateIntervalValidator. Debugging the code it seems that Spring is having trouble instantiating a "metaDataSourceAdvisor" bean somewhere deep in the implementation, but it doesn't seem to have this issue when instantiating a NullOrNotBlankValidator.

I pushed a working version to `placeholder/better-events` but I just checked and the version in this PR is working well in my machine :thinking:

BrunoRosendo:
If the tests are the problem (I didn't quite get it), they can probably be fixed by #80, you should update your branch

ttoino:
> > > @ttoino Any progress on this?
> > 
> > 
> > I'm still getting the same error referring to the DateIntervalValidator. Debugging the code it seems that Spring is having trouble instantiating a "metaDataSourceAdvisor" bean somewhere deep in the implementation, but it doesn't seem to have this issue when instantiating a NullOrNotBlankValidator.
> 
> I pushed a working version to `placeholder/better-events` but I just checked and the version in this PR is working well in my machine thinking

I'm getting the same error on that branch or on this one after rebasing. It is indeed a problem with the tests, but it only happens if I run the whole test suite, running a single test or just the event tests does not produce this error.

BrunoRosendo:
@ttoino I merged the PR fixing the tests into your branch. Please check if everything is ready so we can merge both into develop

ttoino:
> @ttoino I merged the PR fixing the tests into your branch. Please check if everything is ready so we can merge both into develop

I think this is ready

src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/EventDto.kt
```diff
@@ -1,11 +1,15 @@
...
-import java.util.Date
+import pt.up.fe.ni.website.backend.model.embeddable.DateInterval
 
 class EventDto(
     val title: String,
     val description: String,
     val registerUrl: String?,
-    val date: Date
+    val dateInterval: DateInterval,
+    val url: String?,
```
BrunoRosendo: url is the same as registerUrl

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -15,8 +20,24 @@ class Event(
     @field:URL
     var registerUrl: String? = null,
 
+    @Embedded
+    @field:Valid
+    val dateInterval: DateInterval,
+
+    @field:NullOrNotBlank
+    @field:URL
+    val url: String?,
```
BrunoRosendo: same with url, only one is needed

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -147,19 +279,326 @@ internal class EventControllerTest @Autowired constructor(
...
+                    }
+                },
+                requiredFields = mapOf(
+                    "title" to testEvent.title,
+                    "description" to testEvent.description,
+                    "dateInterval" to testEvent.dateInterval,
+                    "url" to testEvent.url,
+                    "location" to testEvent.location,
+                    "category" to testEvent.category,
+                    "thumbnailPath" to testEvent.thumbnailPath
```
BrunoRosendo: Are they all required?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -94,7 +218,11 @@ internal class EventControllerTest @Autowired constructor(
                 requiredFields = mapOf(
                     "title" to testEvent.title,
                     "description" to testEvent.description,
-                    "date" to testEvent.date
+                    "dateInterval" to testEvent.dateInterval,
+                    "url" to testEvent.url,
+                    "location" to testEvent.location,
+                    "category" to testEvent.category,
+                    "thumbnailPath" to testEvent.thumbnailPath
```
BrunoRosendo: Are they all required?

BrunoRosendo (changes requested):
I probably shouldn't haver merged into your branch, the changes got a bit confusing. But now it's done and we can still understand

BrunoRosendo (approved):

DoStini (approved):

PR #59, opened 2022-11-30
Tests/add code coverage

### LINKED ISSUE #37: tests: Add test coverage to the project
Since we have tests after #33, it'd be nice to showcase our coverage in the GitHub CI Actions and also in the README.

As for the Actions, it should possible to do with some kind of flag in `ci.yml`. As for the README, we could consider adding a badge from [Codecov](https://about.codecov.io/) or a similar tool. We should be able to easily integrate it with our Github CI. You can look at multiple projects using this, but here's an example: https://github.com/codecov/example-java-gradle.

It's also possible to integrate this tool with pull requests and it'll tell us how the coverage has increased or decreased. We can consider using this if we think is worth it.

pedronunes19:
Closes #37 

Attempts to add code coverage reports to the project and integrate it with GitHub Actions.

# Review checklist
-   [x] Properly documents API changes in `docs/openapi.yml`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,4 @@ jobs:
         run: ./gradlew test
 
       - name: Upload code coverage to Codecov
-        uses: codecov/codecov-action@v3  
-        with:
-          directory: ./reports/jacoco/
+        uses: codecov/codecov-action@v3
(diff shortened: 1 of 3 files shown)


BrunoRosendo (approved):
Great job! :rocket:

PR #68, opened 2022-12-18
Refactored dto structure to reside in a single package

### LINKED ISSUE #67: global: refactor DTO structure
Until a while ago, all the DTOs we had were a direct mapping between DTO and model, so we had a `Dto` class with useful methods like `create()` and `update()`. However, we quickly realized we need more than that kind of DTO (see discussion in https://github.com/NIAEFEUP/website-niaefeup-backend/pull/54#discussion_r1038903546). The result was a few DTOs residing alongside the controller classes, which is not ideal.

The discussion above proposes the following solution:
```
dto/
     model/
     auth/
     otherSemanticUseCase/
```

This `dto` package should either reside in the global scope of the application or inside the `controller` package, I'll leave that to discussion. It'd also be nice to consider cases of model DTOs that need more fields than the respective model's fields (e.g. creating a project with a list of account IDs to be used as its team).

BrunoRosendo:
closes #67 

I decided to create the single DTO package in the global scope, since DTOs are used both in the controllers and in the services

# Review checklist


bdmendes (approved):
Looks good to me!

MRita443 (approved):

PR #71, opened 2022-12-19
File upload

### LINKED ISSUE #11: assets: investigate storage and link to activities
Since both projects, events, posts and members should be feeded with images, we should investigate how to store them, possibly in Cloudfare CDN, that we have an account for and use in the current website.

DoStini:
Closes #11 

# Review checklist


src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
```diff
@@ -19,5 +20,5 @@ class AccountController(private val service: AccountService) {
     fun getAccountById(@PathVariable id: Long) = service.getAccountById(id)
 
     @PostMapping("/new")
-    fun createAccount(@RequestBody dto: AccountDto) = service.createAccount(dto)
+    fun createAccount(@RequestPart account: AccountDto, @RequestPart image: MultipartFile?) = service.createAccount(account)
```
DoStini: I'm having trouble getting everything together in one single DTO.
In order to support files, we need to use multipart/form-data, and that doesn't work with `@RequestBody`. We would have to use `@ModelAttribute`. However, this annotation seems to not parse the dto correctly and parse some fields incorrectly:

image

The only solution I found was to separate the account dto and the image file upload in two separate form fields, causing the request to look like this:

image


Some references

https://stackoverflow.com/questions/23533355/spring-controller-requestbody-with-file-upload-is-it-possible#answer-31753785

https://stackoverflow.com/questions/61343477/how-to-send-multipart-and-requestbody-together-in-spring-boot-2-x

build.gradle.kts
```diff
@@ -29,11 +31,15 @@ dependencies {
...
     implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+    implementation("com.cloudinary:cloudinary:1.0.14")
     annotationProcessor("org.springframework.boot:spring-boot-configuration-processor")
     implementation("org.springframework.boot:spring-boot-starter-validation:3.0.0")
     developmentOnly("org.springframework.boot:spring-boot-devtools")
     runtimeOnly("com.h2database:h2")
-    testImplementation("org.springframework.boot:spring-boot-starter-test")
+    testImplementation("org.springframework.boot:spring-boot-starter-test") {
+        exclude(group = "org.mockito2", module = "mockito-core")
+    }
```
BrunoRosendo, line 49: Why do we exclude mockito?

build.gradle.kts
```diff
@@ -45,6 +51,30 @@ tasks.withType<KotlinCompile> {
...
+                TestLogEvent.SKIPPED,
+                TestLogEvent.STANDARD_OUT,
+                TestLogEvent.STANDARD_ERROR,
+                TestLogEvent.STANDARD_OUT
+            )
+            exceptionFormat = TestExceptionFormat.FULL
+        }
+        info.events = debug.events
+        info.exceptionFormat = debug.exceptionFormat
+    }
```
BrunoRosendo: What does this change?

src/main/kotlin/pt/up/fe/ni/website/backend/annotations/validation/ValidImage.kt
```diff
@@ -0,0 +1,47 @@
...
+        return true
+    }
+
+    private fun isSupportedContentType(contentType: String?): Boolean {
+        return contentType == "image/png" || contentType == "image/jpg" || contentType == "image/jpeg"
+    }
+
+    private fun isSupportedFileExtension(extension: String?): Boolean {
+        return extension == "png" || extension == "jpg" || extension == "jpeg"
+    }
```
BrunoRosendo: I'll suggest storing them in lists so it's easier to change them. Lmk what you think

src/main/kotlin/pt/up/fe/ni/website/backend/config/upload/UploadConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
+import pt.up.fe.ni.website.backend.util.CloudinaryFileUploader
+import pt.up.fe.ni.website.backend.util.FileUploader
+import pt.up.fe.ni.website.backend.util.StaticFileUploader
+
+@Configuration
+class UploadConfig(
+    private val uploadConfigProperties: UploadConfigProperties,
+
+) : WebMvcConfigurer {
```
BrunoRosendo: Why do we need to extend WebMvcConfigurer? I believe it isn't done in other configs

src/main/kotlin/pt/up/fe/ni/website/backend/config/upload/UploadConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+    @Bean
+    fun fileUploader(): FileUploader {
+        return when (uploadConfigProperties.uploadType) {
+            "cloudinary" -> CloudinaryFileUploader(
+                uploadConfigProperties.cloudinaryBasePath ?: "/",
+                Cloudinary(
+                    uploadConfigProperties.cloudinaryUrl
+                        ?: kotlin.run { throw Error("Cloudinary URL not provided") }
+                )
```
BrunoRosendo: Maybe adding a variable here is more readable and avoids `kotlin.run`

BrunoRosendo: Other constructor params are simpler so I don't mind

src/main/kotlin/pt/up/fe/ni/website/backend/config/upload/UploadConfigProperties.kt
```diff
@@ -0,0 +1,12 @@
+package pt.up.fe.ni.website.backend.config.upload
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties(prefix = "upload")
+class UploadConfigProperties(
+    val uploadType: String?,
```
BrunoRosendo:
```suggestion
    val provider: String?,
```
upload seems redundant, do you think this is more explanatory?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
```diff
@@ -1,23 +1,32 @@
...
     @GetMapping
     fun getAllAccounts() = service.getAllAccounts()
 
     @GetMapping("/{id}")
     fun getAccountById(@PathVariable id: Long) = service.getAccountById(id)
 
-    @PostMapping("/new")
-    fun createAccount(@RequestBody dto: AccountDto) = service.createAccount(dto)
+    @PostMapping("/new", consumes = ["multipart/form-data"])
+    fun createAccount(@RequestPart account: AccountDto, @RequestParam @ValidImage photo: MultipartFile?): Account {
```
BrunoRosendo: #108 needs to implement this as well in edit

BrunoRosendo: Do you think `account` is a better name than `dto`? dto seems less confusing to me

build.gradle.kts
```diff
@@ -45,6 +51,30 @@ tasks.withType<KotlinCompile> {
...
+
+        debug {
+            events = mutableSetOf(
+                TestLogEvent.FAILED,
+                TestLogEvent.PASSED,
+                TestLogEvent.SKIPPED,
+                TestLogEvent.STANDARD_OUT,
+                TestLogEvent.STANDARD_ERROR,
+                TestLogEvent.STANDARD_OUT
+            )
```
ttoino: Is `STANDARD_OUT` supposed to be here twice?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -18,6 +21,10 @@ class AccountService(private val repository: AccountRepository, private val enco
 
         val account = dto.create()
         account.password = encoder.encode(dto.password)
+
+        val fileName = "${UUID.randomUUID()}.${dto.photoFile?.filenameExtension() ?: ""}"
```
BrunoRosendo: We could probably define photoFile as lateInit to get rid of those null checks. Might cause an exception if someone forgets it (and I'm not sure if it's actually a good thing)

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
```diff
@@ -1,23 +1,32 @@
...
 import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.multipart.MultipartFile
+import pt.up.fe.ni.website.backend.annotations.validation.ValidImage
 import pt.up.fe.ni.website.backend.dto.entity.AccountDto
+import pt.up.fe.ni.website.backend.model.Account
 import pt.up.fe.ni.website.backend.service.AccountService
 
 @RestController
 @RequestMapping("/accounts")
+@Validated
```
ttoino, line 24: I'm curious, why is `@Validated` needed here?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
```diff
@@ -40,7 +42,9 @@ class ErrorController : ErrorController {
                 SimpleError(
                     violation.message,
                     violation.propertyPath.toString(),
-                    violation.invalidValue
+                    violation.invalidValue?.let {
+                        if (isSerializable(it)) it else null
+                    }
```
ttoino: Really minor, but I think using `takeIf(::isSerializable)` instead of `let` would be more idiomatic in this case.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
BrunoRosendo: Do you remember the hell we had when fixing members' images? What if we preprend their name before the random id?

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
```diff
@@ -0,0 +1,47 @@
...
+                "transformation" to imageTransformation,
+            )
+        )
+
+        return result["url"]?.toString() ?: ""
+    }
+
+    override fun delete(filePath: String) {
+        println("upload cloud")
+    }
```
ttoino: Is this just a placeholder for now?

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
```diff
@@ -0,0 +1,47 @@
...
+        val file = File("$storePath/$folder/$fileName")
+        file.createNewFile()
+        file.writeBytes(image)
+
+        return "$servePath/$folder/$fileName"
+    }
+
+    override fun delete(filePath: String) {
+        println("delete static")
+    }
```
ttoino: Same as above

src/main/kotlin/pt/up/fe/ni/website/backend/annotations/validation/ValidImage.kt
ttoino: I think this is fine for now but in the future we could investigate a way to to support more image types and maybe a way to automatically convert and optimize them to a format like webp

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
```diff
@@ -0,0 +1,47 @@
+package pt.up.fe.ni.website.backend.util
```
BrunoRosendo: I apologize for my aggression but I firmly believe the utils package can be the most common anti-pattern there is (https://www.yanglinzhao.com/posts/utils-antipattern/). `util` gives no information about what we can find inside it, tends to grow way too much, and most code you usually find in there could be separated in meaningful packages.

In this case, FileUploader seems like a clear service for uploading files so I think you should move these 3 classes there. I would also put them in separate files under `services/upload` for example. Everything else should work the same way w/ the config

ttoino (commented):
This looks really good and paves the way for a lot of features, nice job! Just a few questions here and there but nothing major.

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import com.cloudinary.Cloudinary
+import com.cloudinary.Transformation
+import java.io.File
+
+interface FileUploader {
+    fun upload(folder: String, fileName: String, image: ByteArray): String
```
BrunoRosendo:
```suggestion
    fun uploadImage(folder: String, fileName: String, image: ByteArray): String
```
This is specific for images (since we resize them). There will probably be a regular upload in the future

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
```diff
@@ -0,0 +1,47 @@
...
+                "overwrite" to true,
+                "transformation" to imageTransformation,
+            )
+        )
+
+        return result["url"]?.toString() ?: ""
+    }
+
+    override fun delete(filePath: String) {
+        println("upload cloud")
```
BrunoRosendo: Is this for a future issue? We can use the TODO exception for that

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
```diff
@@ -0,0 +1,47 @@
...
+    override fun upload(folder: String, fileName: String, image: ByteArray): String {
+        val file = File("$storePath/$folder/$fileName")
+        file.createNewFile()
+        file.writeBytes(image)
+
+        return "$servePath/$folder/$fileName"
+    }
+
+    override fun delete(filePath: String) {
+        println("delete static")
```
BrunoRosendo: same as above

src/main/kotlin/pt/up/fe/ni/website/backend/util/MultipartFile.kt
```diff
@@ -0,0 +1,7 @@
+package pt.up.fe.ni.website.backend.util
+
+import org.springframework.web.multipart.MultipartFile
+
+fun MultipartFile.filenameExtension(): String? {
+    return this.originalFilename?.substringAfterLast(".") ?: ""
+}
```
BrunoRosendo: What this returns is not nullable, right? I'd either change to `String` or remove `?: ""`. Maybe the second is more predictable

BrunoRosendo: I think extensions do make sense for a utils package, but we should explain that better. I suggest changing the name to `MultipartFileExtensions` or creating an `extensions` package inside `utils`. Perhaps it's a good idea to go w/ the second approach and also move `annotations` to `utils`. That way, utils is just an aggregation of many utility packages with meaningful names. If you think `utils/` seems too underwhelming for such a task, I've seen `pkg/` used in a lot of GO projects

src/main/kotlin/pt/up/fe/ni/website/backend/util/Serialize.kt
```diff
@@ -0,0 +1,15 @@
...
+fun isSerializable(value: Any): Boolean {
+    return try {
+        val objectMapper = ObjectMapper()
+        objectMapper.writeValueAsString(value)
+        true
+    } catch (err: InvalidDefinitionException) {
+        System.err.println(err)
+        false
+    }
+}
```
BrunoRosendo: Do you think that'll be used outside ErrorController? It doesn't actually check if the value is serializable, it's basically used as a safe guard. I also had this problem in #88, what do you think of my approach?

image

src/main/resources/application.properties
```diff
@@ -25,6 +25,17 @@ auth.public-key=classpath:certs/public.pem
 auth.jwt-access-expiration-minutes=60
 auth.jwt-refresh-expiration-days=7
 
+# File Upload Config
+# Max file size.
+spring.servlet.multipart.max-file-size=500KB
```
BrunoRosendo:
```suggestion
spring.servlet.multipart.max-file-size=500KB
```
redundant imo

src/main/resources/validation_errors.properties
```diff
@@ -1,3 +1,4 @@
 size.min = size must be greater or equal to {min}
 null_or_not_blank.error = must be null or not blank
 date_interval.error = endDate must be after startDate
+files.invalid_image = invalid image format
```
BrunoRosendo:
```suggestion
files.invalid_image = invalid image type (png, jpg or jpeg)
```
Something more informative would be nice

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -157,42 +163,71 @@ class AccountControllerTest @Autowired constructor(
...
+            val file = MockMultipartFile(
+                "photo",
+                "photo.jpeg",
+                MediaType.IMAGE_JPEG_VALUE,
+                "content".toByteArray()
+            )
+
+            val uuid: UUID = UUID.randomUUID()
+            Mockito.mockStatic(UUID::class.java)
+            Mockito.`when`(UUID.randomUUID()).thenReturn(uuid)
```
BrunoRosendo, line 272: Do you know why the `s are used in kotlin? So odd

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -402,21 +427,83 @@ class AccountControllerTest @Autowired constructor(
...
             println("testAccount: ${objectMapper.writeValueAsString(testAccount)}")
-            mockMvc.post("/accounts/new") {
-                contentType = MediaType.APPLICATION_JSON
-                content = testAccount.toJson()
+
+            val accountPart = MockPart("account", testAccount.toJson().toByteArray())
+            accountPart.headers.contentType = MediaType.APPLICATION_JSON
+
+            mockMvc.multipart("/accounts/new") {
+                part(accountPart)
```
BrunoRosendo: This code is everywhere, what if we extend mockMvc with a new method that takes a content string?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -157,42 +163,71 @@ class AccountControllerTest @Autowired constructor(
...
-                        "email" to noWebsite.email,
-                        "password" to noWebsite.password,
-                        "bio" to noWebsite.bio,
-                        "birthDate" to noWebsite.birthDate,
-                        "photoPath" to noWebsite.photoPath,
-                        "linkedin" to noWebsite.linkedin,
-                        "github" to noWebsite.github
-                    )
-                )
+            val accountPart = MockPart("account", noWebsite.toJson().toByteArray())
```
BrunoRosendo: I don't think this is the same. websites will be serialized as `[]`, so the test is useless. That's why it previously used mapOf

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -282,21 +317,6 @@ class AccountControllerTest @Autowired constructor(
...
-            @NestedTest
-            @DisplayName("photoPath")
-            inner class PhotoPathValidation {
-                @BeforeAll
-                fun setParam() {
-                    validationTester.param = "photoPath"
-                }
-
-                @Test
-                fun `should be null or not blank`() = validationTester.isNullOrNotBlank()
```
BrunoRosendo, line 356: Why can't this be tested?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -402,21 +427,83 @@ class AccountControllerTest @Autowired constructor(
...
+            val file = MockMultipartFile(
+                "photo",
+                "photo.pdf",
+                MediaType.IMAGE_JPEG_VALUE,
+                "content".toByteArray()
+            )
+
+            val uuid: UUID = UUID.randomUUID()
+            val mockedSettings = Mockito.mockStatic(UUID::class.java)
+            Mockito.`when`(UUID.randomUUID()).thenReturn(uuid)
```
BrunoRosendo: why do we need mocks here?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -402,21 +427,83 @@ class AccountControllerTest @Autowired constructor(
...
+            val file = MockMultipartFile(
+                "photo",
+                "photo.jpeg",
+                MediaType.APPLICATION_PDF_VALUE,
+                "content".toByteArray()
+            )
+
+            val uuid: UUID = UUID.randomUUID()
+            val mockedSettings = Mockito.mockStatic(UUID::class.java)
+            Mockito.`when`(UUID.randomUUID()).thenReturn(uuid)
```
BrunoRosendo, line 752: Same as above

BrunoRosendo (changes requested):
Great job with this PR! I left a lot of questions and suggestions, but they're just cosmetic. I think this is ready feature-wise 🚀

src/main/kotlin/pt/up/fe/ni/website/backend/config/upload/UploadConfig.kt
DoStini: I don't think i understand. Can you clarify?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
DoStini: Validated is here so that we can use custom validators in the create acccount method, such as `@ValidImage`

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
DoStini: Seems good

src/main/kotlin/pt/up/fe/ni/website/backend/annotations/validation/ValidImage.kt
DoStini: I think cloudinary supports these kinds of trasformations

build.gradle.kts
DoStini: We exclude mockito that comes default with spring boot. We need another version of mockito to use tests with static methods

build.gradle.kts
DoStini: This provides a list of tests and shows why tests fail when exceptions occur. Try it out (or see the pr logs) to see if you like this format

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
DoStini: > We could probably define photoFile as lateInit to get rid of those null checks. Might cause an exception if someone forgets it (and I'm not sure if it's actually a good thing)

I think photofile can be null since the photo is not a required field, so I think this is unnecessary.

> Do you remember the hell we had when fixing members' images? What if we preprend their name before the random id?

Yes this seems like a good idea. I will use email so spaces are not presnt

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
DoStini: Do you think this will ever be necessary? I left it for tought for later

src/main/kotlin/pt/up/fe/ni/website/backend/util/Serialize.kt
DoStini: That seems like a good aproach, I will remove this serialize file

src/main/kotlin/pt/up/fe/ni/website/backend/util/MultipartFile.kt
DoStini: Done

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
DoStini: I redone the repo structure according to your suggestion of using utils with extensions and annotations

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
DoStini: I found this: https://kotlinlang.org/docs/java-interop.html#escaping-for-java-identifiers-that-are-keywords-in-kotlin

Basicly it serves to escape java methods, variables, etc that are keywords in kotlin. It also serves to escape function names(?) like the tests

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
DoStini: We don't include this param in the request

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
DoStini: UUID is a static method that we cannot inject. We could also probably use seeds of random, but I believe this is a more correct way of testing this

build.gradle.kts
BrunoRosendo: I don't see the difference by looking at the logs (failed tests), I might just be blind though

src/main/kotlin/pt/up/fe/ni/website/backend/config/upload/UploadConfig.kt
BrunoRosendo: I meant something like this (form chatgpt, not tested)

````kotlin
fun fileUploader(): FileUploader {
    return if (uploadConfigProperties.provider == "cloudinary") {
        val cloudinaryBaseUrl = uploadConfigProperties.cloudinaryBasePath ?: "/"
        val cloudinaryUrl = uploadConfigProperties.cloudinaryUrl
            ?: throw Error("Cloudinary URL not provided")
        CloudinaryFileUploader(cloudinaryBaseUrl, Cloudinary(cloudinaryUrl))
    } else {
        val staticPath = uploadConfigProperties.staticPath?.let { ResourceUtils.getFile(it).absolutePath } ?: ""
        val staticServe = uploadConfigProperties.staticServe ?: "localhost:8080"
        StaticFileUploader(staticPath, staticServe)
    }
}
````

File: src/main/kotlin/pt/up/fe/ni/website/backend/service/upload/FileUploader.kt [hunk 29]
```diff
@@ -0,0 +1,47 @@
+package pt.up.fe.ni.website.backend.service.upload
```
BrunoRosendo, line 1: I would only create the `upload/` package if the classes live in 3 different files. In this case, it seems weird

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
BrunoRosendo: It's cool to have but not a priority I think. I would leave but I think we should throw the exception, a print might not be readable enough in the logs

src/main/kotlin/pt/up/fe/ni/website/backend/util/MultipartFile.kt
BrunoRosendo: Please check my first comment again, seems weird to return "" by default when the type is nullable

src/main/resources/validation_errors.properties
BrunoRosendo: any decision on this?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
BrunoRosendo: Interesting

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -167,91 +174,72 @@ class AccountControllerTest @Autowired constructor(
...
-                .andDocument(
-                    documentation,
-                    "Create new accounts",
-                    "This endpoint operation creates a new account.",
-                    documentRequestPayload = true
                 )
         }
 
         @Test
-        fun `should create an account with an empty website list`() {
```
BrunoRosendo: Why was this deleted?

BrunoRosendo (changes requested):
I responded to some comments and wrote a question.

The API docs for creating accounts is not well formated. Is this a task for #144 ?

BrunoRosendo:
This now needs to take into account the changes made in #108

src/main/kotlin/pt/up/fe/ni/website/backend/service/upload/FileUploader.kt
DoStini: I think they should be in different files, what do you think?

BrunoRosendo: I agree

src/main/kotlin/pt/up/fe/ni/website/backend/config/upload/UploadConfig.kt
DoStini: I think adding a variable here is unnecessary as it won't add a lot of meaning and won't clean the code that much.
I have removed the run function though

src/main/kotlin/pt/up/fe/ni/website/backend/util/FileUploader.kt
DoStini: It's probably better to simply remove this entirely and open a new issue

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
DoStini: I think the new methodology using a builder pattern is quite clean

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
DoStini: Redone it

src/test/kotlin/pt/up/fe/ni/website/backend/utils/mockmvc/MockMvcMultipartBuilder.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.test.web.servlet.request.RequestPostProcessor
+
+class MockMvcMultipartBuilder(
+    private val mockMvc: MockMvc,
+    uri: String
+) {
+    private val multipart = multipart(uri)
+
+    fun addPart(key: String, data: String): MockMvcMultipartBuilder {
+        val part = MockPart("dto", data.toByteArray())
```
BrunoRosendo: Is this always `"dto"`? `key` doesn't seem to be used

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
BrunoRosendo: I like it too! Good job

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -653,60 +692,127 @@ class AccountControllerTest @Autowired constructor(
...
+                    jsonPath("$.websites[0].url").value(newWebsites[0].url),
+                    jsonPath("$.websites[0].iconPath").value(newWebsites[0].iconPath)
+                )
+//                .andDocument(
+//                    documentation,
+//                    "Update accounts",
+//                    "Update a previously created account, with the exception of its password, using its ID.",
+//                    urlParameters = parameters,
+//                    documentRequestPayload = true
+//            )
```
BrunoRosendo, line 736:
```suggestion
```

BrunoRosendo: There are multiple lines commented, is that intentional?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -35,7 +52,13 @@ class AccountService(private val repository: AccountRepository, private val enco
             throw IllegalArgumentException(ErrorMessages.emailAlreadyExists)
         }
 
+        dto.photoFile?.let {
+            val fileName = photoFilename(dto.email, it)
+            account.photo = fileUploader.uploadImage("profile", fileName, it.bytes)
+        }
+
         val newAccount = dto.update(account)
```
BrunoRosendo, line 60: Is the image kept if not included on `UpdateAccountDto`? I assume it is but wanted to know if you've tested

src/test/kotlin/pt/up/fe/ni/website/backend/utils/mockmvc/MockMvcMultipartBuilder.kt
```diff
@@ -0,0 +1,55 @@
...
+            filename,
+            contentType,
+            content.toByteArray()
+        )
+
+        multipart.file(file)
+        return this
+    }
+
+    fun asPutMethod(): MockMvcMultipartBuilder {
```
BrunoRosendo, line 42: why isn't `asPostMethod` needed?

BrunoRosendo (changes requested):
Great job, just minor details

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
DoStini: Yes, because multipart documentation is not implementaed, but I didn't want to delete the documentation written by other people

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
DoStini: I have a test for this: src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt#749

src/test/kotlin/pt/up/fe/ni/website/backend/utils/mockmvc/MockMvcMultipartBuilder.kt
DoStini: Because `mockmvc.multipart` is hardcoded as a post request

BrunoRosendo (approved):

ttoino (approved):

PR #75, opened 2023-01-05
Feature/post slug

### LINKED ISSUE #25: posts: query a single post by a slug name
Currently, fetching a single post is done using its ID. This is also true for projects and events. Therefore, the page that will be used by the frontend user will be something like `ni.fe.up.pt/posts/134`.
Instead, we can think about using a slang name for a URL like `ni.fe.up.pt/posts/tts-new-release` which is more user friendly. We can also think about doing this for projects, events or other resources where it makes sense.

jamcunha:
Closes #25

This pull request aims to give the option of using a slug instead of an id to request a post from the backend.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
@@ -33,6 +33,6 @@ class PostController(private val service: PostService) {
     @DeleteMapping("/{postId:[0-9]+}")
     fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
 
-    @GetMapping("/{postSlang:[a-z-]+}")
-    fun getPost(@PathVariable postSlang: String) = service.getPostBySlang(postSlang)
+    @GetMapping("/{postSlug:[a-z-]+}")
+    fun getPost(@PathVariable postSlug: String) = service.getPostBySlug(postSlug)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
@@ -18,21 +18,21 @@ class PostController(private val service: PostService) {
     @GetMapping
     fun getAllPosts() = service.getAllPosts()
 
-    @GetMapping("/{postId:[0-9]+}")
+    @GetMapping("/{postId:\\d+}")
     fun getPost(@PathVariable postId: Long) = service.getPostById(postId)
 
     @PostMapping("/new")
     fun createPost(@RequestBody dto: PostDto) = service.createPost(dto)
 
-    @PutMapping("/{postId:[0-9]+}")
+    @PutMapping("/{postId:\\d+}")
     fun updatePost(
         @PathVariable postId: Long,
         @RequestBody dto: PostDto
     ) = service.updatePostById(postId, dto)
 
-    @DeleteMapping("/{postId:[0-9]+}")
+    @DeleteMapping("/{postId:\\d+}")
     fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
 
-    @GetMapping("/{postSlug:[a-z-]+}")
+    @GetMapping("/{postSlug}**")
     fun getPost(@PathVariable postSlug: String) = service.getPostBySlug(postSlug)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
@@ -24,13 +24,13 @@ class PostController(private val service: PostService) {
     @PostMapping("/new")
     fun createPost(@RequestBody dto: PostDto) = service.createPost(dto)
 
-    @PutMapping("/{postId:\\d+}")
+    @PutMapping("/{postId}")
     fun updatePost(
         @PathVariable postId: Long,
         @RequestBody dto: PostDto
     ) = service.updatePostById(postId, dto)
 
-    @DeleteMapping("/{postId:\\d+}")
+    @DeleteMapping("/{postId}")
     fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
 
     @GetMapping("/{postSlug}**")
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
@@ -18,18 +18,21 @@ class PostController(private val service: PostService) {
     @GetMapping
     fun getAllPosts() = service.getAllPosts()
 
-    @GetMapping("/{postId}")
+    @GetMapping("/{postId:[0-9]+}")
     fun getPost(@PathVariable postId: Long) = service.getPostById(postId)
 
     @PostMapping("/new")
     fun createPost(@RequestBody dto: PostDto) = service.createPost(dto)
 
-    @PutMapping("/{postId}")
+    @PutMapping("/{postId:[0-9]+}")
     fun updatePost(
         @PathVariable postId: Long,
         @RequestBody dto: PostDto
     ) = service.updatePostById(postId, dto)
 
-    @DeleteMapping("/{postId}")
+    @DeleteMapping("/{postId:[0-9]+}")
     fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
+
+    @GetMapping("/{postSlang:[a-z-]+}")
+    fun getPost(@PathVariable postSlang: String) = service.getPostBySlang(postSlang)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
@@ -21,6 +21,9 @@ class PostController(private val service: PostService) {
     @GetMapping("/{postId:\\d+}")
     fun getPost(@PathVariable postId: Long) = service.getPostById(postId)
 
+    @GetMapping("/{postSlug}**")
+    fun getPost(@PathVariable postSlug: String) = service.getPostBySlug(postSlug)
+
     @PostMapping("/new")
     fun createPost(@RequestBody dto: PostDto) = service.createPost(dto)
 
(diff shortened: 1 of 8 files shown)


BrunoRosendo (changes requested):
Writing the comment to say that this feature should have some tests. You can take a look at the `PostControllerTest` class and see how other fields are tested

src/test/kotlin/pt/up/fe/ni/website/backend/controller/PostControllerTest.kt
```diff
@@ -303,6 +343,7 @@ internal class PostControllerTest @Autowired constructor(
             )
 
             @Nested
+            @NestedTestConfiguration(NestedTestConfiguration.EnclosingConfiguration.OVERRIDE)
```
BrunoRosendo, line 461: You can simply put this once above `PostControllerTest` class

BrunoRosendo: Please do this in other controller test classes too, since that is the intended behaviour for all tests

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
```diff
@@ -39,5 +40,10 @@ class Post(
     var lastUpdatedAt: Date? = null,
 
     @Id @GeneratedValue
-    val id: Long? = null
+    val id: Long? = null,
+
+    @Column(unique = true)
+    @JsonProperty(required = false)
```
BrunoRosendo: Nothing wrong here, just a cosmetic comment. This is the same as not having the annotation (default behaviour) so we should perhaps remove it, just to be consistent with other optional fields

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
```diff
@@ -18,18 +18,21 @@ class PostController(private val service: PostService) {
...
     fun updatePost(
         @PathVariable postId: Long,
         @RequestBody dto: PostDto
     ) = service.updatePostById(postId, dto)
 
-    @DeleteMapping("/{postId}")
+    @DeleteMapping("/{postId:[0-9]+}")
     fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
+
+    @GetMapping("/{postSlang:[a-z-]+}")
```
BrunoRosendo: With this regex, we cannot create slugs containing numbers (or any characters except lowercase letters). I think we should this endpoint to handle all requests not being handled by `getById`. That way, we always get a semantic error like "no slang found" instead of "invalid endpoint". I can think of two ways of doing this:
- Use a wildcard, `@GetMapping("/{postSlang}**")`, to lower specificity (see [docs](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc-ann-requestmapping-pattern-comparison))
- Negate the regex of the id

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
```diff
@@ -18,18 +18,21 @@ class PostController(private val service: PostService) {
...
 
-    @PutMapping("/{postId}")
+    @PutMapping("/{postId:[0-9]+}")
     fun updatePost(
         @PathVariable postId: Long,
         @RequestBody dto: PostDto
     ) = service.updatePostById(postId, dto)
 
-    @DeleteMapping("/{postId}")
+    @DeleteMapping("/{postId:[0-9]+}")
```
BrunoRosendo: For the same reason as below, I wouldn't include the regex if we don't have ambiguity. If the user introduces an invalid ID, it's less confusing to see "no post with that id" than "invalid endpoint". The users here are always our frontend friends :relaxed:

src/main/kotlin/pt/up/fe/ni/website/backend/model/Post.kt
```diff
@@ -39,5 +40,10 @@ class Post(
...
+    @field:Size(min = Constants.Slang.minSize, max = Constants.Slang.maxSize)
+    val slang: String? = null
```
BrunoRosendo: I believe we can add `@NullOrNotBlank`

BrunoRosendo (changes requested):
Good job! This is working well and we can implement the same feature for activities too! (new issue #76)

One important thing: The issue had a typo and that escalated, it's `slug` instead of `slang` :stuck_out_tongue: Can you please change that?

Also, some suggestions of additional things we might want to test:
- Validations of the slug field (size, isNullOrNotBlank), similar to others (e.g. TitleValidation)
- Verify the error if a post with the same slug is created or edited

src/test/kotlin/pt/up/fe/ni/website/backend/controller/PostControllerTest.kt
BrunoRosendo: I was wrong, this would also nullify `DirtiesContext` in first-level inner classes. keep it as is

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
jamcunha: For this, I kept the regex and if the parameter passed to the mapping isn't just numbers it'll be considered a slug and the error will be "post not found with slug X".

src/test/kotlin/pt/up/fe/ni/website/backend/controller/PostControllerTest.kt
```diff
@@ -104,6 +106,40 @@ internal class PostControllerTest @Autowired constructor(
         }
     }
 
+    @Nested
+    @DisplayName("GET /posts/{postSlug}")
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    inner class GetPostBySlang {
```
BrunoRosendo:
```suggestion
    inner class GetPostBySlug {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/PostControllerTest.kt
```diff
@@ -104,6 +106,40 @@ internal class PostControllerTest @Autowired constructor(
...
+                }
+        }
+
+        @Test
+        fun `should fail if the post does not exist`() {
+            mockMvc.get("/posts/fail-slang").andExpect {
+                status { isNotFound() }
+                content { contentType(MediaType.APPLICATION_JSON) }
+                jsonPath("$.errors.length()") { value(1) }
+                jsonPath("$.errors[0].message") { value("post not found with slug fail-slang") }
```
BrunoRosendo:
```suggestion
            mockMvc.get("/posts/fail-slug").andExpect {
                status { isNotFound() }
                content { contentType(MediaType.APPLICATION_JSON) }
                jsonPath("$.errors.length()") { value(1) }
                jsonPath("$.errors[0].message") { value("post not found with slug fail-slug") }
```

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
BrunoRosendo: That works well for the GET but not for PUT or DELETE, since slugs are not used in those cases. The error will be "invalid endpoint" if you use a string

BrunoRosendo (changes requested):
Besides the typos, just missing some extra tests that might be useful (in the comment above)

src/main/kotlin/pt/up/fe/ni/website/backend/service/PostService.kt
```diff
@@ -14,19 +14,31 @@ class PostService(private val repository: PostRepository) {
...
 
     fun updatePostById(postId: Long, dto: PostDto): Post {
-        val project = getPostById(postId)
-        val newProject = dto.update(project)
-        return repository.save(newProject)
+        val post = getPostById(postId)
+
+        repository.findBySlug(dto.slug)?.let {
+            throw IllegalArgumentException(ErrorMessages.slugAlreadyExists)
+        }
```
BrunoRosendo: If we do this, then we won't be able to update a post if we don't change the slug. This happens because our update is essentially a recreation of the object. We can fix this with a condition:
```suggestion
        repository.findBySlug(dto.slug)?.let {
            if (it.id != post.id) throw IllegalArgumentException(ErrorMessages.slugAlreadyExists)
        }
```

We should have a test for that case (edit without changing slug)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/PostControllerTest.kt
```diff
@@ -104,9 +106,48 @@ internal class PostControllerTest @Autowired constructor(
...
+            repository.save(testPost)
+        }
+
+        @Test
+        fun `should return the post`() {
+            mockMvc.get("/posts/${testPost.slug}")
+                .andExpect {
+                    status { isOk() }
+                    content { contentType(MediaType.APPLICATION_JSON) }
+                    jsonPath("$.title") { value(testPost.title) }
```
BrunoRosendo, line 125: very small comment but maybe compare id too

src/test/kotlin/pt/up/fe/ni/website/backend/controller/PostControllerTest.kt
```diff
@@ -174,22 +236,6 @@ internal class PostControllerTest @Autowired constructor(
...
-                fun setParam() {
-                    validationTester.param = "thumbnailPath"
-                }
-
-                @Test
-                fun `should be required`() = validationTester.isRequired()
-
-                @Test
-                fun `should not be empty`() = validationTester.isNotEmpty()
-            }
```
BrunoRosendo: Why did this get removed?

BrunoRosendo: We could also add a SlugValidation class to test the size and null_or_not_blank annotations

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
```diff
@@ -32,4 +32,7 @@ class PostController(private val service: PostService) {
 
     @DeleteMapping("/{postId}")
     fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
+
+    @GetMapping("/{postSlug}**")
+    fun getPost(@PathVariable postSlug: String) = service.getPostBySlug(postSlug)
```
BrunoRosendo: Also a small detail, but I'd personally put this get next to id get, so we can see them next to each other and rapidly understand why the `\\d´` and `**` modifiers are used

BrunoRosendo (changes requested):

src/test/kotlin/pt/up/fe/ni/website/backend/controller/PostControllerTest.kt
jamcunha: This didn't get removed, I just moved a test that was after validation to be before.

Edit: Seems like I deleted by accident. Changing now.

BrunoRosendo (approved):
LGTM!

coutinho21 (approved):
Good job!

PR #77, opened 2023-01-09
Register for a future event

### LINKED ISSUE #62: events: register for a future event
A new feature for this website is the possibility of registering for future events directly on the event's page (see mockups).
For starters, this could be a simple URL (e.g. google forms) that is only visible when the event's date is in the future.

pedronunes19:
Closes #62  

Adds a `registerUrl` parameter to `Event` which should be a link to a registration platform (most likely google forms).  
Tests were updated accordingly.  

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
@@ -3,6 +3,7 @@ package pt.up.fe.ni.website.backend.model
 import com.fasterxml.jackson.annotation.JsonProperty
 import jakarta.persistence.Entity
 import org.hibernate.validator.constraints.URL
+import pt.up.fe.ni.website.backend.annotations.validation.NullOrNotBlank
 import java.util.Date
 
 @Entity
@@ -10,9 +11,9 @@ class Event(
     title: String,
     description: String,
 
-    @JsonProperty(required = true)
+    @field:NullOrNotBlank
     @field:URL
-    var registerUrl: String,
+    var registerUrl: String? = null,
 
     @JsonProperty(required = true)
     val date: Date,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
```diff
@@ -2,13 +2,18 @@ package pt.up.fe.ni.website.backend.model
...
 import java.util.Date
 
 @Entity
 class Event(
     title: String,
     description: String,
 
+    @JsonProperty(required = true)
+    @field:URL
+    var registerUrl: String,
```
BrunoRosendo: I don't think this should be required because we'll support past events as well (that is being done in #55 )

BrunoRosendo (changes requested):
Good job, just a small comment. After this and #55 are merged we have the minimum basis for events, so we can start thinking about implementing event registrations in the platform. That'll be a more complex topic that still needs discussion so I'll create an issue for that

src/main/kotlin/pt/up/fe/ni/website/backend/model/Event.kt
pedronunes19: I forgot to consider adding a past event. I've made changes so that registerUrl has the same properties as other links used, for example, in the Account model.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -98,6 +101,7 @@ internal class EventControllerTest @Autowired constructor(
                 requiredFields = mapOf(
                     "title" to testEvent.title,
                     "description" to testEvent.description,
+                    "registerUrl" to testEvent.registerUrl,
```
BrunoRosendo:
```suggestion
```
Not required anymore. It doesn't change behaviour but just to avoid confusion

BrunoRosendo (changes requested):
Working as expected, just a small cosmetic detail

BrunoRosendo (approved):

limwa (approved):

PR #80, opened 2023-01-11
Adds custom test annotations for simplicity and consistency

BrunoRosendo:
For context, @jamcunha was having some problems when using the `@BeforeAll` feature in nested test classes. We then found out that happened because of a misconfiguration where the context was being discarded but the test instance was not over (using class life cycle). We found the solution for that but I realized the annotations in the test classes were getting way too verbose and hard to understand.

Therefore, I tried to find a solution to simplify test writing and avoid confusing when copying random annotations from somewhere else. I liked the solution of creating custom annotations serving as a composition of meta-annotations. This PR also ensures test consistency by using the following pattern:
- Base classes for integration tests are marked with `@ControllerTest`, which configures SpringBootTest, MVC, DB, and marks the dirty context after a test class.
- Inner classes testing endpoints (which always happens in our tests) are marked with `@EndpointTest`. This marks the class with `@Nested` and changes the test instance to `PER_CLASS` life cycle, which ensures we can use features such as `@BeforeAll`.
- Additional inner test classes are marked with `@NestedTest`. This is where @jamcunha's solution was moved to. Here, the configuration is reset (w/ `@NestedTestConfiguration(NestedTestConfiguration.EnclosingConfiguration.OVERRIDE)`) to make sure the context is not invalidated in these classes (dirty context). This avoids the bug mentioned above and improves performance since unnecessary context reloads are not done anymore.

# Review checklist

### DIFF
--- a/src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
+++ b/src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
@@ -3,11 +3,8 @@ package pt.up.fe.ni.website.backend.controller
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeAll
-import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.DisplayName
-import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
-import org.junit.jupiter.api.TestInstance
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.MediaType
 import org.springframework.test.web.servlet.MockMvc
@@ -19,6 +16,8 @@ import pt.up.fe.ni.website.backend.repository.AccountRepository
 import pt.up.fe.ni.website.backend.utils.TestUtils
 import pt.up.fe.ni.website.backend.utils.ValidationTester
 import pt.up.fe.ni.website.backend.utils.annotations.ControllerTest
+import pt.up.fe.ni.website.backend.utils.annotations.EndpointTest
+import pt.up.fe.ni.website.backend.utils.annotations.NestedTest
 import java.util.Calendar
 import java.util.Date
 import pt.up.fe.ni.website.backend.model.constants.AccountConstants as Constants
@@ -43,9 +42,8 @@ class AccountControllerTest @Autowired constructor(
         )
     )
 
-    @Nested
+    @EndpointTest
     @DisplayName("GET /accounts")
-    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
     inner class GetAllAccounts {
         private val testAccounts = listOf(
             testAccount,
@@ -78,9 +76,8 @@ class AccountControllerTest @Autowired constructor(
         }
     }
 
-    @Nested
+    @EndpointTest
     @DisplayName("GET /accounts/{id}")
-    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
     inner class GetAccount {
         @BeforeAll
         fun addAccount() {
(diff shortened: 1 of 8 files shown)


src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -49,9 +42,8 @@ class AccountControllerTest @Autowired constructor(
         )
     )
 
-    @Nested
+    @EndpointTest
     @DisplayName("GET /accounts")
```
BrunoRosendo, line 46: I spent some time trying to merge these two annotations but didn't succeed. Some help would be appreciated, you can see the solution I tried in the thread I opened: https://stackoverflow.com/questions/75077738/kotlin-aliasfor-not-overriding-attribute-from-displayname

MRita443 (approved):

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
bdmendes: I spent some time looking at this myself and it seems to me that your solution is correct and does work when the composed annotation attribute is not present: https://chrysanthium.com/spring-annotation-composition
However, `DisplayName` has no default value for its `value` field, which may be causing the issue. Not sure why this works this way but found a possibly similar case: https://stackoverflow.com/questions/69280914/spring-aliasfor-not-working-for-profile-annotation

bdmendes (commented):
Awesome quality of life improvement!
I'm not fond of splitting "util" annotations and "non util" ones in the test directory. Both of them are utils to me. As such, I would suggest moving the annotations directory to the utils:
```
tests/
     controller/
          ...
     utils/
          annotations/
               composites/
                    ....
               validation/
                    ....
```

BrunoRosendo:
> Awesome quality of life improvement! I'm not fond of splitting "util" annotations and "non util" ones in the test directory. Both of them are utils to me. As such, I would suggest moving the annotations directory to the utils:
> 
> ```
> tests/
>      controller/
>           ...
>      utils/
>           annotations/
>                composites/
>                     ....
>                validation/
>                     ....
> ```

One is test on annotations (e.g. NullOrNotBlankTest) and another is annotations used for testing. I can see the confusion but the first package contains tests (not utils or annotations). Maybe we could change the names to something else?

MRita443 (changes requested):
An error occurs while testing the project update together with other tests:

HV000221: An error occurred while extracting values in value extractor org.hibernate.validator.internal.engine.valueextraction.ListValueExtractor.

Cause org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.hibernate.validator.internal.constraintvalidators.bv.EmailValidator': BeanPostProcessor before instantiation of bean failed

BrunoRosendo:
> 

From what I investigated, the exception was caused by uninitialized beans regarding Account validation. The error can be fixed on your side by removing the `@Validated` annotations from the teamMembers field since it's only a relationship so it shouldn't need to be validated.

However, I don't understand why this error only happens after executing the `CreateProject` test class or why it happens at all. I thought this would be a problem in cases where we have nested entities that need to be validated, like customWebsites in the Account entity. I tried creating a simple update endpoint and a test similar to the one that was failing in your case. To my surprise, there were no errors whatsoever.

I do not understand what's causing the exceptions in that case (maybe because the relationship doesn't cascade or the entity is already persisted? But that doesn't seem to make much sense). Either way, I don't think this will cause us issues whenever we need to test nested validation so I think it's safe to keep it like this. It might be a weird edge case regarding SpringContext

BrunoRosendo:
By the way, if you want an extra read to understand Spring's ApplicationContext https://www.baeldung.com/spring-application-context

This is what we reset in each endpoint test to get a clean state of db/app

MRita443 (approved):

ttoino (approved):
LGTM!

PR #86, opened 2023-01-21
Technologies field

### LINKED ISSUE #63: projects: add technologies
Projects should have a list of technologies they use (e.g. Kotlin, Spring, Tailwind, etc.).

In a first simple approach, this could be a simple list of strings. We can later think of creating a dedicated model that can be used to compare project technologies and event categories.

coutinho21:
Closes #63 

Added the field for the technologies of a project.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/ProjectDto.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/ProjectDto.kt
@@ -5,5 +5,6 @@ import pt.up.fe.ni.website.backend.model.Project
 class ProjectDto(
     val title: String,
     val description: String,
-    val isArchived: Boolean = false
+    val isArchived: Boolean = false,
+    val technologies: List<String>
 ) : EntityDto<Project>()
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/ProjectDto.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/ProjectDto.kt
@@ -6,5 +6,5 @@ class ProjectDto(
     val title: String,
     val description: String,
     val isArchived: Boolean = false,
-    val technologies: List<String>
+    val technologies: List<String> = emptyList()
 ) : EntityDto<Project>()
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
@@ -7,5 +7,6 @@ class Project(
     title: String,
     description: String,
     var isArchived: Boolean = false,
+    technologies: List<String>,
     id: Long? = null
 ) : Activity(title, description, id)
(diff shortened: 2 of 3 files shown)


src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
```diff
@@ -7,5 +7,6 @@ class Project(
     title: String,
     description: String,
     var isArchived: Boolean = false,
+    technologies: List<String>,
```
BrunoRosendo:
```suggestion
    val technologies: List<String>,
```
Without val this'd be a useless parameter without creating the respective field. Read more in https://www.baeldung.com/kotlin/constructors#primary-constructor

src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/ProjectDto.kt
```diff
@@ -5,5 +5,6 @@ import pt.up.fe.ni.website.backend.model.Project
 class ProjectDto(
     val title: String,
     val description: String,
-    val isArchived: Boolean = false
+    val isArchived: Boolean = false,
+    val technologies: List<String>
```
BrunoRosendo:
```suggestion
    val technologies: List<String> = emptyList()
```
It doesn't need to be required in the dto, just default to empty

BrunoRosendo (changes requested):
Tests should be passing after those changes

BrunoRosendo:
image
It might be a good idea to add something like this in GetProject, CreateProject and similar situations

src/main/kotlin/pt/up/fe/ni/website/backend/model/Project.kt
coutinho21: I tried that before but got this error, therefore i decided to remove it 

image

coutinho21: Error disappeared 🤷‍♂️

BrunoRosendo: It's always a good idea to research errors/warning instead of trying to fix them blindly. Your IDE plugin is probably referring to a previous version of Hibernate ORM (the one who maps the models to the DB) that didn't support conversion of simple arrays and lists. Instead, you would have to define a separate table (e.g. `@ElementCollection`) or a converter (transform list to string and vice versa).

The ORM is now capable of treating simple lists as a "basic plural" and maps it to just one column, which seems appropriate for this case. Read "JDBC array type support" at https://in.relation.to/2022/06/24/hibernate-orm-61-features/

BrunoRosendo: This threads explores alternative ways, before that feature was released https://stackoverflow.com/questions/21059451/cannot-declare-list-property-in-the-jpa-entity-class-it-says-basic-attribute

bdmendes (approved):
Good work!

BrunoRosendo (approved):
lgtm!

PR #88, opened 2023-01-26
Feature/generations

### LINKED ISSUE #42: accounts: implement generations

BrunoRosendo:
closes #42 

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
@@ -34,5 +34,8 @@ class PostController(private val service: PostService) {
     ) = service.updatePostById(postId, dto)
 
     @DeleteMapping("/{postId}")
-    fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
+    fun deletePost(@PathVariable postId: Long): Map<String, String> {
+        service.deletePostById(postId)
+        return emptyMap()
+    }
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -9,7 +9,7 @@ import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import pt.up.fe.ni.website.backend.dto.entity.ProjectDto
-import pt.up.fe.ni.website.backend.service.ProjectService
+import pt.up.fe.ni.website.backend.service.activity.ProjectService
 
 @RestController
 @RequestMapping("/projects")
@@ -36,7 +36,7 @@ class ProjectController(private val service: ProjectService) {
     @PutMapping("/{id}")
     fun updatePostById(
         @PathVariable id: Long,
-        @RequestBody dto: ProjectDto
+        @RequestBody dto: ProjectDto,
     ) = service.updateProjectById(id, dto)
 
     @PutMapping("/{id}/archive")
@@ -48,12 +48,12 @@ class ProjectController(private val service: ProjectService) {
     @PutMapping("/{idProject}/addTeamMember/{idAccount}")
     fun addTeamMemberById(
         @PathVariable idProject: Long,
-        @PathVariable idAccount: Long
+        @PathVariable idAccount: Long,
     ) = service.addTeamMemberById(idProject, idAccount)
 
     @PutMapping("/{idProject}/removeTeamMember/{idAccount}")
     fun removeTeamMemberById(
         @PathVariable idProject: Long,
-        @PathVariable idAccount: Long
+        @PathVariable idAccount: Long,
     ) = service.removeTeamMemberById(idProject, idAccount)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -36,7 +36,7 @@ class ProjectController(private val service: ProjectService) {
     @PutMapping("/{id}")
     fun updatePostById(
         @PathVariable id: Long,
-        @RequestBody dto: ProjectDto,
+        @RequestBody dto: ProjectDto
     ) = service.updateProjectById(id, dto)
 
     @PutMapping("/{id}/archive")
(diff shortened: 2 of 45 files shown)


BrunoRosendo:
I'm having some trouble regarding deserialization of bi-directional relationships. I created this issue on jackson-databind's repo and I'm waiting for a reply https://github.com/FasterXML/jackson-databind/issues/3760

BrunoRosendo:
Update: I found a solution for the previous problem https://github.com/FasterXML/jackson-module-kotlin/issues/294

BrunoRosendo:
I'm trying to use `@OrderColumn` annotation to maintain the order of the roles but I'm having a few problems with lazy initialization... I'll try again later but I'll consider leaving it to another issue since this PR is already very big

BrunoRosendo:
> I'm trying to use `@OrderColumn` annotation to maintain the order of the roles but I'm having a few problems with lazy initialization... I'll try again later but I'll consider leaving it to another issue since this PR is already very big

I solved the issue after fixing the incorrect configuration of bi-directional relationships. See the sources if you need to:
- https://rogue-modron.blogspot.com/2019/03/mappedby-ordercolumn-hibernate.html
- https://www.youtube.com/watch?v=jCYonZey5dY

I'm just missing the tests now

BrunoRosendo:
This is ready for review. This PR ended up touching on A LOT of concepts so I'd appreciate a careful review

Among many things, I had to refactor ActivityService even further. I think there might be a better solution that I didn't find

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -1,16 +1,18 @@
...
 import pt.up.fe.ni.website.backend.model.Activity
 import pt.up.fe.ni.website.backend.repository.ActivityRepository
+import pt.up.fe.ni.website.backend.service.AccountService
+import pt.up.fe.ni.website.backend.service.ErrorMessages
 
 @Service
-abstract class ActivityService<T : Activity>(
-    private val repository: ActivityRepository<T>,
-    private val accountService: AccountService
+abstract class AbstractActivityService<T : Activity>(
```
BrunoRosendo, line 11: I'm unsure if this is the best way to fix the problem... I needed a way to instantiate an ActivityService of Activities. I'd appreciate suggestions for a better pattern

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
```diff
@@ -34,5 +34,8 @@ class PostController(private val service: PostService) {
     ) = service.updatePostById(postId, dto)
 
     @DeleteMapping("/{postId}")
-    fun deletePost(@PathVariable postId: Long) = service.deletePostById(postId)
+    fun deletePost(@PathVariable postId: Long): Map<String, String> {
+        service.deletePostById(postId)
+        return emptyMap()
+    }
```
Sirze01, line 40: I agree with this switch. Having the response bodies here seems to be clearer. I think we should keep the replies consistent across endpoints, and the delete operation on the events endpoint still uses the map created in the service. We should change it too.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -25,12 +25,15 @@ class ProjectController(private val service: ProjectService) {
     fun createNewProject(@RequestBody dto: ProjectDto) = service.createProject(dto)
 
     @DeleteMapping("/{id}")
-    fun deleteProjectById(@PathVariable id: Long) = service.deleteProjectById(id)
+    fun deleteProjectById(@PathVariable id: Long): Map<String, String> {
+        service.deleteProjectById(id)
+        return emptyMap()
+    }
```
Sirze01, line 34: Related to the comment above

Sirze01 (approved):
Repositories translating the method names to database queries internally is really powerful and interesting. It took me some time to figure out what was happening.

Couldn't think of a better way to instantiate the activity service. I think it is OK for now.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
Sirze01: Forgot to reference it... Here it is:

https://github.com/NIAEFEUP/website-niaefeup-backend/blob/a9698c365c51587d9381b34bca6bb02158dcf65e/src/main/kotlin/pt/up/fe/ni/website/backend/controller/EventController.kt#L28-L31

And

https://github.com/NIAEFEUP/website-niaefeup-backend/blob/a9698c365c51587d9381b34bca6bb02158dcf65e/src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/EventService.kt#L46-L54

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
```diff
@@ -26,15 +32,22 @@ class Role(
...
-    @JoinColumn
-    @ManyToMany
-    var accounts: List<@Valid Account> = emptyList(),
-
-    @JoinColumn
-    @OneToMany
-    var associatedActivities: List<@Valid PerActivityRole> = emptyList(),
+    @ManyToMany(mappedBy = "roles")
+    @JsonIgnore // TODO: Decide if we want to return accounts (or IDs) by default
+    @OnDelete(action = OnDeleteAction.CASCADE) // Remove relationship, since this is the non-owner side
```
LuisDuarte1, line 37: Using delete cascade, in this case, doesn't mean that deleting a Role will delete the accounts associated with it?

LuisDuarte1 (commented):
Looks pretty good to me 🚀 , I just have a question about the `@OnDelete` annotation and it's function on the context below.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
BrunoRosendo: This confused me at first too and I actually had that behaviour incorrectly when I first implemented this. That happened because I was using `CascadeType.REMOVE` which is different from `OnDeleteAction.CASCADE`. The `@OnDelete` will delete the table associated with the relationship instead of the account. I believe I wrote a test for that scenario.

This was an article that helped me understand this: https://rogerkeays.com/jpa-cascadetype-remove-vs-hibernate-ondelete

src/main/kotlin/pt/up/fe/ni/website/backend/controller/PostController.kt
BrunoRosendo: Thanks! I changed it

src/main/kotlin/pt/up/fe/ni/website/backend/model/Role.kt
LuisDuarte1: Ok, that makes sense, since this is a many to many relationship, i didn't notice that at first ahah.

LuisDuarte1 (approved):
:rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket:

PR #89, opened 2023-01-27
Feature/team members

MRita443:
Closes #41

Adds a list of team members to Activity's model and includes endpoints for adding and removing team members

# Review checklist
-   [ ] Properly documents API changes in `docs/openapi.yml`
-   [x] Contains enough appropriate tests
-   [x] Behavior is as expected
-   [x] Clean, well structured code

Issue #84, opened 2023-01-13
accounts: recover password


BrunoRosendo:
We should have a way to recover the password if the member forgot it. A way to implement this is by using a JWT token to validate the recovery request, similar to what we have in NIJobs.

DoStini:
What are we gonna use for this? Should we mimic the way we do this in NIJobs or do we want to explore different approaches?

BrunoRosendo:
I think the NIJobs way works well since we're already using JWT for authentication

PR #91, opened 2023-01-27
Change account password

### LINKED ISSUE #85: accounts: change password
We should have a way for the members to change their passwords. This would be a simple endpoint that checks authentication and takes the old and new passwords (the old one for extra verification). The rest would be similar to the register mechanism

coutinho21:
Closes #85 

The objective is to add a new endpoint to change an account's password.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
@@ -22,6 +22,6 @@ class AccountController(private val service: AccountService) {
     @PostMapping("/new")
     fun createAccount(@RequestBody dto: AccountDto) = service.createAccount(dto)
 
-    @PostMapping("/{id}/change-password")
+    @PostMapping("/changePassword/{id}")
     fun changePassword(@PathVariable id: Long, @RequestBody dto: ChangePassDto) = service.changePassword(id, dto)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
@@ -6,6 +6,7 @@ import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
+import pt.up.fe.ni.website.backend.dto.auth.ChangePassDto
 import pt.up.fe.ni.website.backend.dto.entity.AccountDto
 import pt.up.fe.ni.website.backend.service.AccountService
 
@@ -20,4 +21,7 @@ class AccountController(private val service: AccountService) {
 
     @PostMapping("/new")
     fun createAccount(@RequestBody dto: AccountDto) = service.createAccount(dto)
+
+    @PostMapping("/{id}/change-password")
+    fun changePassword(@PathVariable id: Long, @RequestBody dto: ChangePassDto) = service.changePassword(id, dto)
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
@@ -23,5 +23,8 @@ class AccountController(private val service: AccountService) {
     fun createAccount(@RequestBody dto: AccountDto) = service.createAccount(dto)
 
     @PostMapping("/changePassword/{id}")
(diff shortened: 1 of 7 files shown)


DoStini:
We should add some sort of 2 factor authentication to this (by email for example) for security reasons. I'm not sure if this is possible right now or if we should let this for a future issue

BrunoRosendo:
> We should add some sort of 2 factor authentication to this (by email for example) for security reasons. I'm not sure if this is possible right now or if we should let this for a future issue

Is that something usual when changing passwords with the user logged in? It definitely needs to be done in #84 but I'm not sure of the normal flow of a password change

BrunoRosendo:
@coutinho21 Is this ready for review? If so, please remove it from draft

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
```diff
@@ -11,6 +11,8 @@ object ErrorMessages {
 
     const val expiredRefreshToken = "refresh token has expired"
 
+    const val incorrectPassword = "incorrect password"
```
bdmendes: Does this make sense considering we already have `invalid credentials`?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -419,6 +421,53 @@ class AccountControllerTest @Autowired constructor(
...
+            "https://github.com",
+            listOf(
+                CustomWebsite("https://test-website.com", "https://test-website.com/logo.png")
+            ),
+            emptyList()
+        )
+        @BeforeEach
+        fun addAccount() {
+            repository.save(testAccountToo)
+        }
```
bdmendes: I realize you are creating a new account with the password encrypted to save directly in the DB. Ideally, you could find a way to clone the `testAccount` and encrypt the password, inside the `addAccount` method. If not possible, at least make the variable name clearer

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -26,4 +27,15 @@ class AccountService(private val repository: AccountRepository, private val enco
 
     fun getAccountByEmail(email: String): Account = repository.findByEmail(email)
         ?: throw NoSuchElementException(ErrorMessages.emailNotFound(email))
+
+    fun changePassword(id: Long, dto: ChangePassDto): Map<String, String> {
```
bdmendes: Does this need to return a `Map`?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
```diff
@@ -20,4 +21,7 @@ class AccountController(private val service: AccountService) {
 
     @PostMapping("/new")
     fun createAccount(@RequestBody dto: AccountDto) = service.createAccount(dto)
+
+    @PostMapping("/{id}/change-password")
```
bdmendes: Since #89 we are using camelCase for endpoint names. I would prefer `changePassword/{id}`

bdmendes (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
coutinho21: You're right

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
coutinho21: Yes, otherwise the response is a 500 error saying "No acceptable representation". I put it with an empty map for consistency like others class do.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
coutinho21: Looks cleaner to me too

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
coutinho21: Is this a possible approach? 
https://medium.com/@gugabernardo/here-is-the-easiest-way-to-deep-copy-an-object-in-kotlin-cf200c2130b
I don't know if using more dependencies is going to be optimal

BrunoRosendo: https://stackoverflow.com/questions/49903859/deep-copy-using-jackson-string-or-jsonnode

coutinho21:
Closing this pr due to rebasing issues, starting a new branch as well.

PR #94, opened 2023-01-31
Feature/activities slug

### LINKED ISSUE #76: activities: get project/event by slug name
This is the same feature currently being done for posts (see #75 ). It would be much better to display `ni.fe.up.pt/projects/nijobs` instead of the ugly ID `ni.fe.up.pt/projects/2`, so we should store their unique slug names (customizable by the user) and be able to retrieve the activities using them.

Note: We decided it's best to leave the automatic `title -> slug` conversion to the frontend since titles and names might not be unique and the user can correct the slug in the form.

jamcunha:
Closes #76 

Similar to #75, I added an optional slug to the activities.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/service/ProjectService.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/service/ProjectService.kt
@@ -15,6 +15,10 @@ class ProjectService(
     fun getAllProjects(): List<Project> = repository.findAll().toList()
 
     fun createProject(dto: ProjectDto): Project {
+        repository.findBySlug(dto.slug)?.let {
+            throw IllegalArgumentException(ErrorMessages.slugAlreadyExists)
+        }
+
         val project = dto.create()
 
         dto.teamMembersIds?.forEach {
@@ -28,8 +32,16 @@ class ProjectService(
     fun getProjectById(id: Long): Project = repository.findByIdOrNull(id)
         ?: throw NoSuchElementException(ErrorMessages.projectNotFound(id))
 
+    fun getProjectBySlug(projectSlug: String): Project = repository.findBySlug(projectSlug)
+        ?: throw NoSuchElementException(ErrorMessages.projectNotFound(projectSlug))
+
     fun updateProjectById(id: Long, dto: ProjectDto): Project {
         val project = getProjectById(id)
+
+        repository.findBySlug(dto.slug)?.let {
+            if(it.id != project.id) throw IllegalArgumentException(ErrorMessages.slugAlreadyExists)
+        }
+
         val newProject = dto.update(project)
         newProject.apply {
             teamMembers.clear()
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/service/ProjectService.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/service/ProjectService.kt
@@ -39,7 +39,7 @@ class ProjectService(
         val project = getProjectById(id)
 
         repository.findBySlug(dto.slug)?.let {
-            if(it.id != project.id) throw IllegalArgumentException(ErrorMessages.slugAlreadyExists)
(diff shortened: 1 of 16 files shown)


src/main/kotlin/pt/up/fe/ni/website/backend/service/ProjectService.kt
```diff
@@ -12,15 +12,27 @@ class ProjectService(private val repository: ProjectRepository) {
...
 
+    fun getProjectBySlug(projectSlug: String): Project = repository.findBySlug(projectSlug)
+        ?: throw NoSuchElementException(ErrorMessages.projectNotFound(projectSlug))
+
     fun updateProjectById(id: Long, dto: ProjectDto): Project {
         val project = getProjectById(id)
+
+        repository.findBySlug(dto.slug)?.let {
+            if (it.id != project.id) throw IllegalArgumentException(ErrorMessages.slugAlreadyExists)
+        }
```
BrunoRosendo, line 43: #55 needs to take this into consideration for the events

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -101,11 +137,13 @@ internal class EventControllerTest @Autowired constructor(
                 requiredFields = mapOf(
                     "title" to testEvent.title,
                     "description" to testEvent.description,
-                    "date" to testEvent.date
+                    "date" to testEvent.date,
+                    "slug" to testEvent.slug
```
BrunoRosendo, line 342: Do we need this? Slug is not required

BrunoRosendo (changes requested):
Good job! But aren't we missing the getBySlug endpoint for events?

BrunoRosendo (approved):
Looks good! You just need to rebase it with latest changes in develop

BrunoRosendo (approved):

bdmendes (commented):
Good job! I have a small suggestion. Since both projects and events services and repositories are accessing the slug in the exact same way, what do you think of introducing an abstract `ActivityRepository` (and service)?

MRita443:
> Good job! I have a small suggestion. Since both projects and events services and repositories are accessing the slug in the exact same way, what do you think of introducing an abstract `ActivityRepository` (and service)?

I implemented a similar solution in #89, check it out to see if it's what you suggested!

bdmendes:
> > Good job! I have a small suggestion. Since both projects and events services and repositories are accessing the slug in the exact same way, what do you think of introducing an abstract `ActivityRepository` (and service)?
> 
> I implemented a similar solution in #89, check it out to see if it's what you suggested!

I like it! I would just make it a bit more cohesive by using inheritance.

BrunoRosendo:
> > > Good job! I have a small suggestion. Since both projects and events services and repositories are accessing the slug in the exact same way, what do you think of introducing an abstract `ActivityRepository` (and service)?
> > 
> > 
> > I implemented a similar solution in #89, check it out to see if it's what you suggested!
> 
> I like it! I would just make it a bit more cohesive by using inheritance.

Maybe we should wait for #89 merge then. Inheritance might be trickier but it'd be nice to look at an implementation of that

bdmendes (commented):
Looks good to me. However, I thought that the slug was intended to be a small internal representation of the title (perhaps with some parts of the date and/or description) and should be generated automatically upon creating the activity. Would not that be a better approach than the optional one?

jamcunha:
> Looks good to me. However, I thought that the slug was intended to be a small internal representation of the title (perhaps with some parts of the date and/or description) and should be generated automatically upon creating the activity. Would not that be a better approach than the optional one?

We agreed in #75 that the slug would be generated in the frontend, giving the option to write another slug if wanted.

bdmendes:
Awesome, let's get this merged then

bdmendes (approved):

PR #95, opened 2023-02-02
Replaced @DirtiesContext with manual reset of H2 database

BrunoRosendo:
Since `@DirtiesContext` logic kept bringing issues in the tests, I replaced it with a listener that automatically truncates every table from the H2 test database. Additionally, I had to deactivate an optimization on the entity manager to ensure ID consistency.

Here are the 2 primary sources that inspired this implementation:
- https://better-coding.com/springboot-resetting-h2-database-before-each-integration-test/
- https://stackoverflow.com/questions/43586360/how-to-reset-hibernate-sequence-generators

This means that the DB will be reset in EVERY single test, as opposed to only between endpoint tests like before. There was no need for an `@EndpointTest` in this implementation since it's the same as the `@NestedTest`, so I removed it.

# Review checklist


src/test/kotlin/pt/up/fe/ni/website/backend/utils/listeners/DbCleanupListener.kt
```diff
@@ -0,0 +1,99 @@
...
+
+class DbCleanupListener : TestExecutionListener {
+    companion object {
+        private const val SCHEMA_NAME = "PUBLIC"
+        private const val DB_NAME = "H2"
+    }
+
+    override fun beforeTestMethod(testContext: TestContext) {
+        super.beforeTestMethod(testContext)
+        println("Cleaning up database...")
```
DoStini, line 22: Does this show up in the test log?

BrunoRosendo: It does

ttoino (approved):
LGTM!

src/test/kotlin/pt/up/fe/ni/website/backend/utils/listeners/DbCleanupListener.kt
BrunoRosendo: Should I change anything or can we merge? This is an important PR @DoStini

DoStini: Sorry for the delay, it might be a bit cluttered but we can merge

DoStini (approved):

PR #96, opened 2023-02-03
Simplified error message

### LINKED ISSUE #90: accounts: fix websites error
When creating a new account, if the websites field is missing, this error shows up:

image

This error should be changed to something simpler and easier to understand.

jamcunha:
Closes #90 

Simplified error message when "websites" field is missing.

# Review checklist

### DIFF
--- a/src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
+++ b/src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
@@ -142,6 +142,35 @@ class AccountControllerTest @Autowired constructor(
                 jsonPath("$.websites[0].iconPath") { value(testAccount.websites[0].iconPath) }
             }
         }
+        @Test
+        fun `should create an account with a empty website list`() {
+            val noWebsite = Account(
+                "Test Account",
+                "no_website@email.com",
+                "test_password",
+                "This is a test account",
+                TestUtils.createDate(2001, Calendar.JULY, 28),
+                "https://test-photo.com",
+                "https://linkedin.com",
+                "https://github.com"
+            )
+
+            mockMvc.post("/accounts/new") {
+                contentType = MediaType.APPLICATION_JSON
+                content = noWebsite.toJson()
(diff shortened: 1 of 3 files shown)


BrunoRosendo (commented):
What if instead of throwing an error, we just changed the behaviour and added the default value of []? We just need to change the dto to `List<CustomWebsiteDto> = emptyList()` (and maybe the model, although it doesn't really matter)

Edit: We should also write a test to make sure it's working as expected

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -142,6 +142,35 @@ class AccountControllerTest @Autowired constructor(
                 jsonPath("$.websites[0].iconPath") { value(testAccount.websites[0].iconPath) }
             }
         }
+        @Test
+        fun `should create an account with a empty website list`() {
```
```suggestion
        fun `should create an account with an empty website list`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -142,6 +142,35 @@ class AccountControllerTest @Autowired constructor(
...
+                "This is a test account",
+                TestUtils.createDate(2001, Calendar.JULY, 28),
+                "https://test-photo.com",
+                "https://linkedin.com",
+                "https://github.com"
+            )
+
+            mockMvc.post("/accounts/new") {
+                contentType = MediaType.APPLICATION_JSON
+                content = noWebsite.toJson()
```
BrunoRosendo: This doesn't actually test what you want. `Account` has an empty list of websites so `noWebsite.toJson()` will also have an empty list, instead of a missing value. The best way to test this might be by using `mapOf()` or `AccountDto` for example

BrunoRosendo (changes requested):
It works as expected but the test needs a fix

BrunoRosendo (approved):

coutinho21 (approved):
lgtm

PR #108, opened 2023-02-10
Feature/accounts edit and delete

### LINKED ISSUE #93: accounts: edit and delete
Currently, there are no endpoints to edit or delete accounts. We should create them, similar to what's being done with entities like posts or projects

CarlosMealha:
Closes #93 

Adds endpoints to edit and delete accounts 

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/dto/account/UpdateAccountDto.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/dto/account/UpdateAccountDto.kt
@@ -1,0 +1,15 @@
+package pt.up.fe.ni.website.backend.dto.account
+
+import java.util.Date
+import pt.up.fe.ni.website.backend.dto.entity.CustomWebsiteDto
+
+class UpdateAccountDto(
+    val email: String,
+    val name: String,
+    val bio: String?,
+    val birthDate: Date?,
+    val photoPath: String?,
+    val linkedin: String?,
+    val github: String?,
+    val websites: List<CustomWebsiteDto>?
+)
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
@@ -45,11 +45,12 @@ class AccountService(private val repository: AccountRepository, private val enco
         repository.save(account)
     }
 
-    fun deleteAccountById(id: Long) {
+    fun deleteAccountById(id: Long): Map<String, String> {
         if (!repository.existsById(id)) {
             throw NoSuchElementException(ErrorMessages.accountNotFound(id))
         }
 
         repository.deleteById(id)
+        return emptyMap()
     }
 }
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
@@ -3,9 +3,9 @@ package pt.up.fe.ni.website.backend.service
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.security.crypto.password.PasswordEncoder
 import org.springframework.stereotype.Service
-import pt.up.fe.ni.website.backend.dto.account.UpdateAccountDto
 import pt.up.fe.ni.website.backend.dto.auth.ChangePasswordDto
-import pt.up.fe.ni.website.backend.dto.entity.AccountDto
+import pt.up.fe.ni.website.backend.dto.entity.account.CreateAccountDto
+import pt.up.fe.ni.website.backend.dto.entity.account.UpdateAccountDto
 import pt.up.fe.ni.website.backend.model.Account
 import pt.up.fe.ni.website.backend.repository.AccountRepository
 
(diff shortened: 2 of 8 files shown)


CarlosMealha:
Can someone tell me how I can work around the error of using wildcard imports, because in these two cases ```
import jakarta.persistence.*
import org.junit.jupiter.api.*
```, when I individually import, it automatically gives a fix for `*`

**Comment by CarlosMealha** [comment]
> Can someone tell me how I can work around the error of using wildcard imports, because in these two cases `import jakarta.persistence.* import org.junit.jupiter.api.*`, when I individually import, it automatically gives a fix for `*`

So I removed the IDE flag that blocked wildcard imports and added a comment to the README file.

**Comment by BrunoRosendo** [comment]
There are still some linting problems (you can fix them automatically, check readme or just format with the plugin). You also need to rebase now

**Comment by CarlosMealha** [comment]
> There are still some linting problems (you can fix them automatically, check readme or just format with the plugin). You also need to rebase now

Yes I know that, I'm currently testing the edit endpoint and then I will fix that

**Comment by gwealm** [comment]
Things seem to be working from an high level perspective. Can you explain again what the problem was @CarlosMealha?
Just linting issues?

**Comment by CarlosMealha** [comment]
> Things seem to be working from an high level. Can you explain again what the problem was @CarlosMealha?
> Just linting issues?

Yes everything is working, just need to finish testing the edit endpoint, at the moment one of the atributtes is not being changed to the expected one.

**Review by BrunoRosendo — CHANGES_REQUESTED** [review]
I don't think it makes sense to allow for password change here. We already have an endpoint for changing passwords and that operation required sending the old password as well.

I see 2 options:
- Remove the password field from this endpoint (new dto for example)
- Port the password change logic into this endpoint. I'm not the biggest fan of this approach, since the logic is different from a normal update, but I understand its advantages.

What do you think?

File: src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt [hunk 1]
```diff
@@ -27,6 +28,28 @@ class AccountService(private val repository: AccountRepository, private val enco
...
+            dto.email,
+            account.password,
+            dto.name,
+            dto.bio,
+            dto.birthDate,
+            dto.photoPath,
+            dto.linkedin,
+            dto.github,
+            dto.websites
+        )
```
BrunoRosendo: There should be no need to do this, UpdateAccountDto can extend EntityDto

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -38,4 +61,13 @@ class AccountService(private val repository: AccountRepository, private val enco
...
     }
+
+    fun deleteAccountById(id: Long): Map<String, String> {
+        if (!repository.existsById(id)) {
+            throw NoSuchElementException(ErrorMessages.accountNotFound(id))
+        }
+
+        repository.deleteById(id)
+        return emptyMap()
+    }
```
BrunoRosendo: We shouldn't return emptyMap() in the service (we changed this recently). Services should not handle API responses, that's for controllers

src/main/kotlin/pt/up/fe/ni/website/backend/dto/account/UpdateAccountDto.kt
```diff
@@ -0,0 +1,15 @@
+package pt.up.fe.ni.website.backend.dto.account
+
+import java.util.Date
+import pt.up.fe.ni.website.backend.dto.entity.CustomWebsiteDto
+
+class UpdateAccountDto(
```
BrunoRosendo: It might be confusing to have AccountDto when it's only used for creating it. I suggest changing it to CreateAccountDto (similarly to update)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -573,6 +578,224 @@ class AccountControllerTest @Autowired constructor(
...
+            Assertions.assertEquals(newBio, updatedAccount.bio)
+            Assertions.assertEquals(newBirthDate.toJson(), updatedAccount.birthDate.toJson())
+            Assertions.assertEquals(newPhotoPath, updatedAccount.photoPath)
+            Assertions.assertEquals(newLinkedin, updatedAccount.linkedin)
+            Assertions.assertEquals(newWebsites[0].url, updatedAccount.websites[0].url)
+            Assertions.assertEquals(newWebsites[0].iconPath, updatedAccount.websites[0].iconPath)
+        }
+
+        @Test
+        fun `should fail if the project does not exist`() {
```
```suggestion
        fun `should fail if the account does not exist`() {
```

BrunoRosendo (changes requested):

MRita443:
Ever since my latest commit, I have been getting test failures on both `/posts/{postSlug}` and `/posts/{postId}`. I haven't changed these tests, and the errors I am getting are due to `lastUpdatedAt` being an hour earlier than expected. For example:
`JSON path "$.lastUpdatedAt" expected:<05-04-2023 15:07:09> but was:<05-04-2023 14:07:09>`. Can anyone tell me if this is normal and why this seems to be happening?

BrunoRosendo:
> Ever since my latest commit, I have been getting test failures on both `/posts/{postSlug}` and `/posts/{postId}`. I haven't changed these tests, and the errors I am getting are due to `lastUpdatedAt` being an hour earlier than expected. For example: `JSON path "$.lastUpdatedAt" expected:<05-04-2023 15:07:09> but was:<05-04-2023 14:07:09>`. Can anyone tell me if this is normal and why this seems to be happening?

That seems to be a local problem related to the daylight saving time that changed last week. Can you try putting this setting in application.properties and see if it changes? `spring.jackson.time-zone=UTC`

BrunoRosendo:
@MRita443 Actually, you might want to try it in the `toJson()` extension function at the end of the test file, depending on whether the timezone is wrong in the application (`but was`) or the test (`expected`)

BrunoRosendo (approved):
Good job!

LuisDuarte1 (approved):
LGTM 🚀🚀 🚀 🚀 🚀 🚀 🚀

PR #112, opened 2023-02-15
Add code coverage to ci artifact

### LINKED ISSUE #111: devops: add code coverage to test report
Possibly, via a gradle plugin.

ttoino:
Closes #111.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,7 +55,7 @@ tasks.jacocoTestReport {
     dependsOn(tasks.test) // tests are required to run before generating the report
     reports {
         xml.required.set(true)
-        csv.required.set(false)
-        html.required.set(false)
+        csv.required.set(true)
+        html.required.set(true)
     }
 }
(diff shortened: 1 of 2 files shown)


pedronunes19 (approved):

build.gradle.kts
```diff
@@ -55,7 +55,7 @@ tasks.jacocoTestReport {
     dependsOn(tasks.test) // tests are required to run before generating the report
     reports {
         xml.required.set(true)
-        csv.required.set(false)
-        html.required.set(false)
+        csv.required.set(true)
+        html.required.set(true)
```
BrunoRosendo, line 59: Why do we need all those reports? Just to have more options?

BrunoRosendo (approved):
Seems good, I just have a small question so I'm approving

build.gradle.kts
ttoino: We don't really need all those, I just turned them on because it's the default settings

PR #113, opened 2023-02-20
Feature/recover password

### LINKED ISSUE #84: accounts: recover password

jamcunha:
Closes #84 

# Review checklist


src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -27,6 +28,13 @@ class AuthController(val authService: AuthService) {
         return mapOf("access_token" to accessToken)
     }
 
+    @PostMapping("/recoverPassword/{id}")
+    fun generateRecoveryToken(@PathVariable id: Long): Map<String, String> {
+        val recoveryToken = authService.generateRecoveryToken(id)
+        // TODO: Change URL Later
+        return mapOf("recovery_url" to "localhost:8080/accounts/recoverPassword/$recoveryToken")
```
BrunoRosendo: We'll probably need to set some property for the URL

BrunoRosendo (changes requested):
The logic looks good overall. I didn't do a careful review since this is just a draft

BrunoRosendo:
@jamcunha Is this ready for review or do you still have some doubts? We can talk about it in the next meeting or in 1on1

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -901,6 +904,80 @@ class AccountControllerTest @Autowired constructor(
...
+                                )
+                            )
+                    ).andExpectAll(
+                        status().isOk()
+                    ).andDocumentCustomRequestSchema(
+                        documentation,
+                        passwordRecoveryPayload,
+                        "Recover password",
+                        "Update the password of an account using a recovery token.",
+                        // urlParameters = parameters,
```
jamcunha: It is pretty much ready for review. There is a doubt here because it errors when I document the URL parameters.

BrunoRosendo: Hmm lets see it in the next meeting

bdmendes (commented):
The code looks good to me. To improve security a bit, we can look into serializing the current password into the jwt to make sure it is only used once: https://stackoverflow.com/a/54865104

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
```diff
@@ -29,6 +30,10 @@ class AccountController(private val service: AccountService) {
     @GetMapping("/{id}")
     fun getAccountById(@PathVariable id: Long) = service.getAccountById(id)
 
+    @PutMapping("/recoverPassword/{recoveryToken}")
+    fun recoverPassword(@RequestBody dto: PassRecoveryDto, @PathVariable recoveryToken: String) =
+        service.recoverPassword(recoveryToken, dto)
+
     @PostMapping("/changePassword/{id}")
     fun changePassword(@PathVariable id: Long, @RequestBody dto: ChangePasswordDto): Map<String, String> {
         service.changePassword(id, dto)
```
BrunoRosendo: This can be considered in #124 but I think the methods for changing or resetting a password should use the same method (probably post since we're referring to password in the URL)

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -28,6 +33,13 @@ class AuthController(val authService: AuthService) {
         return mapOf("access_token" to accessToken)
     }
 
+    @PostMapping("/recoverPassword/{id}")
+    fun generateRecoveryToken(@PathVariable id: Long): Map<String, String> {
+        val recoveryToken = authService.generateRecoveryToken(id)
+        // TODO: Change URL Later
+        return mapOf("recovery_url" to "$backendUrl/accounts/recoverPassword/$recoveryToken")
```
BrunoRosendo: Should this be the backend URL? CC @DoStini but I think the URL users click takes them to a frontend page (w/ input for the new password) that then calls the recover password endpoint. Therefore, instead of using the backendUrl we could have a property for the recovery URL specifically

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/PassRecoveryDto.kt
```diff
@@ -0,0 +1,5 @@
+package pt.up.fe.ni.website.backend.dto.auth
+
+data class PassRecoveryDto(
```
BrunoRosendo: small detail but shortening the word doesn't do much and could cause confusion when searching for the class. It's best to be consistent with `ChangePasswordDto` or `recoverPassword` for example

Therefore, `PasswordRecoveryDto`

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -901,6 +905,83 @@ class AccountControllerTest @Autowired constructor(
...
+            )
+        )
+
+        @BeforeEach
+        fun setup() {
+            repository.save(testAccount)
+        }
+
+        @Test
+        fun `should update the password`() {
```
BrunoRosendo: It'd be nice to test if the password hash really changed

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -901,6 +905,83 @@ class AccountControllerTest @Autowired constructor(
...
+                    ).andDocumentCustomRequestSchema(
+                        documentation,
+                        passwordRecoveryPayload,
+                        "Recover password",
+                        "Update the password of an account using a recovery token.",
+                        urlParameters = parameters,
+                        documentRequestPayload = true
+                    )
+                }
+        }
```
BrunoRosendo: This test is pretty good for the current version but it won't work once we update to sending emails. Don't make the changes in this PR but please create an issue with the following info:
- Link to this PR and explain that email service wasn't ready.
- Explain what needs to be added.
- Explain how we should test the new behaviour. We can think about using mocks to check whether the right functions are being called and not actually send emails. You can check how it's being done in NIJobs as I believe it's pretty well implemented: https://github.com/NIAEFEUP/nijobs-be/blob/4ab4f74789c0568ecfa8a4f3c4c618ae13df5ad3/test/end-to-end/auth.js#L243

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/PassRecoveryDto.kt
```diff
@@ -0,0 +1,5 @@
...
+    val password: String
```
BrunoRosendo: I'd consider putting a constraint to enforce a minimum size to the password, as we have in the account model. Currently, I can reset to an easy password like `1`

BrunoRosendo (changes requested):
Very good job with this implementation! I left some comments with small suggestions and doubts

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
jamcunha: I should rename that to `apiUrl` as it better describes my thoughts when implementing it. I didn't consider being a frontend page because we didn't have one, but certainly is the intention when implementing sending emails.

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/PassRecoveryDto.kt
jamcunha: While doing this, I noticed that `ChangePasswordDto` has the same problem. Should I fix it in this PR?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
BrunoRosendo: Is there any point leaving it with the backend URL? Why not implement the right logic and only send the email in the future issue?

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/PassRecoveryDto.kt
BrunoRosendo: Yes that would be great!

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
jamcunha: Makes sense. Do we already have the frontend page?

BrunoRosendo: We don't but just put a config property as a placeholder

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/ChangePasswordDto.kt
```diff
@@ -1,6 +1,12 @@
 package pt.up.fe.ni.website.backend.dto.auth
 
+import jakarta.validation.constraints.Size
+import pt.up.fe.ni.website.backend.model.constants.AccountConstants as Constants
+
 data class ChangePasswordDto(
+    @field:Size(min = Constants.Password.minSize, max = Constants.Password.maxSize)
```
BrunoRosendo: Checking the old password seems redundant. It probably won't make a difference but this could block changing passwords if we ever populate the db directly (again, shouldn't happen but ***could*** happen)

BrunoRosendo (approved):
Good job, I left a small comment but I'm pre-approving. It'd be nice if you could review your commits too, maybe squash a few

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
BrunoRosendo: @jamcunha please don't forget to create the issue :)

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -59,29 +59,43 @@ class AuthService(
...
             .claim("scope", scope)
-            .build()
+
+        if (usePasswordHash) {
+            claimsBuilder.claim("passwordHash", account.password)
+        }
+
+        val claims = claimsBuilder.build()
         return jwtEncoder.encode(JwtEncoderParameters.from(claims)).tokenValue
     }
```
jamcunha: I don't know if this is what it is supposed to do regarding the password serialization mentioned in this PR, but it seems right to me.

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/ChangePasswordDto.kt
jamcunha: Makes sense, removed it.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
jamcunha: Done. Please let me know if more information is needed.
Issue: #171

BrunoRosendo (changes requested):
Regarding the latest change, where do you check to see if the hash present in the jwt is the same as the one in the db?

BrunoRosendo:
> Regarding the latest change, where do you check to see if the hash present in the jwt is the same as the one in the db?

nevermind this comment, I was looking at the wrong diff

BrunoRosendo (changes requested):
can you create 1/2 tests for the latest change? Maybe 1 where a user tries to use the same recovery link twice and another one just to check for missing hash (if the last one isn't too hard)

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
```diff
@@ -11,6 +11,12 @@ object ErrorMessages {
 
     const val expiredRefreshToken = "refresh token has expired"
 
+    const val invalidRecoveryToken = "invalid password recovery token"
+
+    const val expiredRecoveryToken = "password recovery token has expired"
+
+    const val missingHashClaim = "password hash claim is missing"
```
BrunoRosendo: I'm thinking if we should return a general `invalid recovery token` error instead of specifying it. This may be a detail we want to omit from users, wdyt @DoStini ?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -57,26 +57,45 @@ class AuthService(
         return generateAccessToken(account)
     }
 
+    fun generateRecoveryToken(id: Long): String {
```
BrunoRosendo:
```suggestion
    private fun generateRecoveryToken(id: Long): String {
```

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
```diff
@@ -29,8 +31,20 @@ class AccountController(private val service: AccountService) {
...
+    ) =
+        service.recoverPassword(recoveryToken, dto)
+
     @PostMapping("/changePassword/{id}")
-    fun changePassword(@PathVariable id: Long, @RequestBody dto: ChangePasswordDto): Map<String, String> {
+    fun changePassword(
+        @Valid @RequestBody
+        dto: ChangePasswordDto,
+        @PathVariable id: Long
+    ): Map<String, String> {
```
DoStini: Regarding the routing of things, I think they should be the following:
- account/password/recover
- account/password/update

Do you need the account ID here? Shouldn't we be using the session account id?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -14,6 +16,9 @@ import pt.up.fe.ni.website.backend.service.AuthService
 @RestController
 @RequestMapping("/auth")
 class AuthController(val authService: AuthService) {
+    @field:Value("\${page.recover-password}")
+    private lateinit var recoverPasswordPage: String
+
```
DoStini, line 25: Curious, what is this for?

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
jamcunha: I did it that way because it would only error if it's not implemented well in the code (forgetting to set `usePasswordHash` to true and use it). But if it eventually occurs with the user, it makes sense to omit it.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -28,6 +33,13 @@ class AuthController(val authService: AuthService) {
         return mapOf("access_token" to accessToken)
     }
 
+    @PostMapping("/recoverPassword/{id}")
+    fun generateRecoveryToken(@PathVariable id: Long): Map<String, String> {
```
DoStini: Seems weird to me that this is in a different controller.

I suggest either moving everything to auth controller or to account controller.

Besides that, there are two requests with a similar URL `recoverPassword`.

I suggest the following routes in AuthController:
post -> password/recovery with email in body
post -> password/recovery/{token}/confirm
post -> password/update

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -28,6 +33,13 @@ class AuthController(val authService: AuthService) {
...
+        val recoveryToken = authService.generateRecoveryToken(id)
```
DoStini: Users don't know the account ID, the users should provide an email.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -28,6 +33,13 @@ class AuthController(val authService: AuthService) {
...
+        // TODO: Change to email service
+        return mapOf("recovery_url" to "$recoverPasswordPage/$recoveryToken")
```
DoStini: For security reasons, users should not be able to know if a email is registered in a website, so the request should always be `Ok 200`

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/ChangePasswordDto.kt
```diff
@@ -1,6 +1,11 @@
 package pt.up.fe.ni.website.backend.dto.auth
 
+import jakarta.validation.constraints.Size
+import pt.up.fe.ni.website.backend.model.constants.AccountConstants as Constants
+
 data class ChangePasswordDto(
     val oldPassword: String,
+
+    @field:Size(min = Constants.Password.minSize, max = Constants.Password.maxSize)
```
DoStini, line 9: Passwords don't need a maximum size in my opinion

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/PasswordRecoveryDto.kt
```diff
@@ -0,0 +1,9 @@
+package pt.up.fe.ni.website.backend.dto.auth
+
+import jakarta.validation.constraints.Size
+import pt.up.fe.ni.website.backend.model.constants.AccountConstants as Constants
+
+data class PasswordRecoveryDto(
+    @field:Size(min = Constants.Password.minSize, max = Constants.Password.maxSize)
```
DoStini, line 7: Passwords don't need a maximum size in my opinion

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -65,6 +70,29 @@ class AccountService(
     fun getAccountByEmail(email: String): Account = repository.findByEmail(email)
         ?: throw NoSuchElementException(ErrorMessages.emailNotFound(email))
 
+    fun recoverPassword(recoveryToken: String, dto: PasswordRecoveryDto): Account {
+        val jwt =
+            try {
+                jwtDecoder.decode(recoveryToken)
+            } catch (e: Exception) {
+                throw InvalidBearerTokenException(ErrorMessages.invalidRecoveryToken)
```
DoStini: If this creates an error response in json, please consider removing it. Just use a return statement to stop the flow and maybe the log the error. This is to ensure that no information about what emails are stored in our databse

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
DoStini: Is this regarding password changing only 1 time with the link?

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
DoStini: I was just reading this now, I've mentioned in my previous comments that we should omit these kinds of errors to the users so that information on who uses the website doesn't get leaked. Always return ok even if the account doesn't exist

DoStini (changes requested):

DoStini:
Overall looking good, just left some comments to improve the RESTiness of the API and privacy concerning issues 🚀

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
jamcunha: To use the property in `application.properties` file.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
jamcunha: Yes

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
bdmendes: Indeed, one can get the authenticated account via a method in the auth service

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -65,6 +70,29 @@ class AccountService(
...
+            }
+        if (jwt.expiresAt?.isBefore(Instant.now()) != false) {
+            throw InvalidBearerTokenException(ErrorMessages.expiredRecoveryToken)
+        }
+        val account = getAccountByEmail(jwt.subject)
+
+        val tokenPasswordHash = jwt.getClaim<String>("passwordHash")
+            ?: throw InvalidBearerTokenException(ErrorMessages.missingHashClaim)
```
bdmendes: What if the hash is not in the payload (since we have that option)? You can simply let it be null and skip the next check if that is the case.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
jamcunha: If it's private it will not work, or am I missing something?

jamcunha:
Just changed the recovery URL request to be provided with the email instead of the id and removed errors related to invalid tokens, logging them instead.

> Seems weird to me that this is in a different controller.
> 
> I suggest either moving everything to auth controller or to account controller.
> 
> Besides that, there are two requests with a similar URL `recoverPassword`.
> 
> I suggest the following routes in AuthController: post -> password/recovery with email in body post -> password/recovery/{token}/confirm post -> password/update

The tests may be lacking because I didn't want to test more without making these changes.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
BrunoRosendo: you're right of course, my bad

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/ChangePasswordDto.kt
BrunoRosendo: @DoStini the maximum size is also used when creating an account. I believe this was done to prevent ridiculously long passwords but, now that I think about it, we only store the hash so there's no point. I believe it's fine to use the default value of Integer.MAX_VALUE here and in the account model

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
BrunoRosendo: @DoStini do we always assume a password change was successful even if the URL is malformed? It seems weird for the user to think the password changed when reality it didn't

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -17,8 +19,9 @@ import pt.up.fe.ni.website.backend.utils.extensions.filenameExtension
 class AccountService(
     private val repository: AccountRepository,
     private val encoder: PasswordEncoder,
+    private val jwtDecoder: JwtDecoder,
     private val fileUploader: FileUploader
-) {
+) : Logging {
```
BrunoRosendo: Is the logger being used?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -57,26 +59,84 @@ class AuthService(
...
+        if (jwt.expiresAt?.isBefore(Instant.now()) != false) {
+            throw InvalidBearerTokenException(ErrorMessages.expiredRecoveryToken)
+        }
+        val account = accountService.getAccountByEmail(jwt.subject)
+
+        val tokenPasswordHash =
+            try {
+                jwt.getClaim<String>("passwordHash")
+            } catch (e: Exception) {
+                // throw InvalidBearerTokenException(ErrorMessages.invalidRecoveryToken)
```
BrunoRosendo: please remove these comments when it's done

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -57,26 +59,84 @@ class AuthService(
...
+        )
+    }
+
+    fun confirmRecoveryToken(recoveryToken: String, dto: PasswordRecoveryConfirmDto): Account? {
+        val jwt =
+            try {
+                jwtDecoder.decode(recoveryToken)
+            } catch (e: Exception) {
+                // throw InvalidBearerTokenException(ErrorMessages.invalidRecoveryToken)
+                logger.error(ErrorMessages.invalidRecoveryToken)
```
BrunoRosendo: I don't think error makes sense here. logger.error is usually for server-side or unexpected errors, this is just a malformed request. Imo info or warn makes more sense

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -57,26 +59,84 @@ class AuthService(
...
+            }
+
+        if (account.password != tokenPasswordHash) {
+            // throw InvalidBearerTokenException(ErrorMessages.invalidRecoveryToken)
+            logger.error(ErrorMessages.invalidRecoveryToken)
+            return null
+        }
+
+        account.password = passwordEncoder.encode(dto.password)
+        return accountService.updateAccount(account)
```
BrunoRosendo: @DoStini this is related to my previous comment but, if we want to hide invalid tokens then the user can distinguish between null and the account object. In that case, we should always return empty map on success

BrunoRosendo (changes requested):
small changes, Im also waiting for @DoStini response on my comment

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
DoStini: You are right, I missunderstood the use of this function. Only errors in the act of `requesting password change` should be hidden from the user. If the URL is malformed or expired whenever they are setting the new password, the user should be warned about it!

This should go back to how it was. Sorry @jamcunha !!!

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
DoStini: Now that I'm reading this again, it makes sense to have an `invalid recovery token` error whenever we try to set a new password.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -14,6 +16,9 @@ import pt.up.fe.ni.website.backend.service.AuthService
 @RestController
 @RequestMapping("/auth")
 class AuthController(val authService: AuthService) {
+    @field:Value("\${page.recover-password}")
+    private lateinit var recoverPasswordPage: String
```
DoStini, line 24: Shouldn't this be passed by config and config properties or is it bloat to use? @BrunoRosendo @bdmendes

BrunoRosendo: What do you mean? It is passed by the config property

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
jamcunha: But if we remove that, how can we tell if it is or isn't successful?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -57,26 +58,69 @@ class AuthService(
...
+            throw InvalidBearerTokenException(ErrorMessages.expiredRecoveryToken)
+        }
+        val account = accountService.getAccountByEmail(jwt.subject)
+
+        val tokenPasswordHash = jwt.getClaim<String>("passwordHash")
+            ?: throw InvalidBearerTokenException(ErrorMessages.invalidRecoveryToken)
+
+        if (account.password != tokenPasswordHash) {
+            throw InvalidBearerTokenException(ErrorMessages.invalidRecoveryToken)
+        }
```
BrunoRosendo: maybe expired makes more sense because the token was already used

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -57,26 +58,69 @@ class AuthService(
         return generateAccessToken(account)
     }
 
+    fun generateRecoveryToken(email: String): String? {
```
BrunoRosendo, line 83:
```suggestion
    fun generateRecoveryToken(email: String): String {
```
Correct me if I'm wrong but I believe it can't be null

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
```diff
@@ -180,6 +193,268 @@ class AuthControllerTest @Autowired constructor(
...
+    inner class RecoverPasswordRequest {
+        var documentation: ModelDocumentation = PayloadRecoverPassword()
+
+        @BeforeEach
+        fun setup() {
+            repository.save(testAccount)
+        }
+
+        @Test
+        fun `should fail if email is not found`() {
```
BrunoRosendo: After what @DoStini said, I don't think this should be failing (just 200 and do nothing)

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadAuth.kt
```diff
@@ -56,6 +56,26 @@ class PayloadAuthRefresh : ModelDocumentation(
...
+            "Email of the account",
+            JsonFieldType.STRING,
+            isInResponse = false
+        ),
+        DocumentedJSONField(
+            "recovery_url",
+            "URL to recover the password",
+            JsonFieldType.STRING,
+            isInRequest = false,
+            optional = true // change this when email service is implemented
```
BrunoRosendo:
```suggestion
            optional = true // TODO change this when email service is implemented
```
for easy track

BrunoRosendo (changes requested):
Small comments, besides the one still unresolved

Also, can you find a way to test the uncovered (red) lines? It'd be awesome if it's not too much trouble

image

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
jamcunha: Seems better for me too.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
jamcunha: It should. I mistakenly removed it after reading something about failing when confirming the token and assumed fail all for some reason.

jamcunha:
> Small comments, besides the one still unresolved
> 
> Also, can you find a way to test the uncovered (red) lines? It'd be awesome if it's not too much trouble
> 
> image

I'm having problems trying to test those two lines because when testing token expiration, I try to create a new token with all the claims of the previous token but modifying the `expireAt` to `Instant.now()` but it's caught in an exception when trying to decode in `confirmRecoveryToken`. I think it's the same for `passwordHash` claim.

jamcunha:
For now, I only changed the error handling in the recovery logic so as not to mess up with the tests. After some inputs, I will change the handling for other jwt logic.

BrunoRosendo (approved):
Needs a rebase but I'm pre-approving

LuisDuarte1 (approved):
LGTM 🚀, it looks clean to me, just solve the conflicts below 😅

LuisDuarte1:
@jamcunha can you fix the conflicts please?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -28,6 +36,21 @@ class AuthController(val authService: AuthService) {
         return mapOf("access_token" to accessToken)
     }
 
+    @PostMapping("/password/recovery")
+    fun generateRecoveryToken(@RequestBody recoveryRequestDto: PasswordRecoveryRequestDto): Map<String, String> {
+        val recoveryToken = authService.generateRecoveryToken(recoveryRequestDto.email)
+            ?: return emptyMap()
+        // TODO: Change to email service
+        return mapOf("recovery_url" to "$recoverPasswordPage/$recoveryToken/confirm")
```
DoStini: Make this return OK all the time and comment on what is supposed to go on an email. If someone wants to test this right away, one can simply print the result. It is safer to leave the correct behavior already implemented.

```suggestion
        val recoveryToken = authService.generateRecoveryToken(recoveryRequestDto.email)
            ?: return emptyMap()
        // TODO:  Add email service with payload "$recoverPasswordPage/$recoveryToken/
        
        return ""
```

DoStini: I'm not sure about the `return ""`, but it should always be status 200. 

Also document on why this is done for security reasons, in order to not leak what are the emails associated with our service

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -44,39 +45,69 @@ class AuthService(
...
-        val jwt =
-            try {
-                jwtDecoder.decode(refreshToken)
-            } catch (e: Exception) {
-                throw InvalidBearerTokenException(ErrorMessages.invalidRefreshToken)
-            }
-        if (jwt.expiresAt?.isBefore(Instant.now()) != false) {
-            throw InvalidBearerTokenException(ErrorMessages.expiredRefreshToken)
-        }
+        val jwt = jwtDecoder.decode(refreshToken)
```
DoStini: Why the changes here?

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -44,39 +45,69 @@ class AuthService(
...
         val account = accountService.getAccountByEmail(jwt.subject)
         return generateAccessToken(account)
     }
 
+    fun generateRecoveryToken(email: String): String? {
+        val account = try {
+            accountService.getAccountByEmail(email)
+        } catch (e: Exception) {
+            return null
```
DoStini, line 88: If the error is not the expected one (account not found), log the error so that developers can more easily find the problem

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
```diff
@@ -44,39 +45,69 @@ class AuthService(
...
+        val tokenPasswordHash = jwt.getClaim<String>("passwordHash")
+            ?: throw InvalidBearerTokenException(ErrorMessages.invalidToken)
+
+        if (account.password != tokenPasswordHash) {
+            throw InvalidBearerTokenException(ErrorMessages.invalidToken)
+        }
+
+        account.password = passwordEncoder.encode(dto.password)
+        return accountService.updateAccount(account)
+    }
```
DoStini: This function also updates the account password, so I don't think this is the best name. I would either change the function name a little but, or extract the update logic and use this as an auxiliary function to validate the token

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
```diff
@@ -180,6 +201,367 @@ class AuthControllerTest @Autowired constructor(
...
+                    status().isOk(),
+                    jsonPath("$.recovery_url").exists()
+                ).andDocument(
+                    documentation,
+                    "Recover password",
+                    "This endpoint operation allows the recovery of the password of an account, " +
+                        "sending an email with a link to reset the password.",
+                    documentRequestPayload = true
+                )
+        }
```
DoStini, line 258: As I explained before, this is not the expected behavior. Both cases will return empty and ok status.

What you should do is test the request returns empty and ok both times, but also to test if the first situation does not call the generate token function. I think the easiest way would be by mocking JwtClaimsSet and verifying if some of those functions do not get called.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -28,6 +36,21 @@ class AuthController(val authService: AuthService) {
         return mapOf("access_token" to accessToken)
     }
 
+    @PostMapping("/password/recovery")
+    fun generateRecoveryToken(@RequestBody recoveryRequestDto: PasswordRecoveryRequestDto): Map<String, String> {
+        val recoveryToken = authService.generateRecoveryToken(recoveryRequestDto.email)
+            ?: return emptyMap()
```
DoStini, line 45:
```suggestion
        val recoveryToken = authService.generateRecoveryToken(recoveryRequestDto.email)
        
```

DoStini (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
jamcunha: I removed the try catch block and handled the exception in the  `ErrorController` and the token expiration validation is already handled in the `decode()` method.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
jamcunha: Couldn't we have it return the recovery link for now so we can test the feature and remove it when the email service is complete?

If I'm not wrong it always returns status 200.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
jamcunha: If someone uses an email that's not related to an account it would trigger the exception. I don't think we should log this.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
jamcunha: Same as above, this could be as is for now and changed when the email service is implemented.
Can leave a comment to remind that.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
rubuy-74: As it is, it always returns status 200.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
rubuy-74: Not needed as this is resolved in the comment below

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AuthControllerTest.kt
rubuy-74: This is going to be as is for now and changed in the email service integration issue #171,

DoStini (approved):
Honestly I dont even know what I requested to change but if my homie @rubuy-74 asks me to allow the merge of this beautiful pull request, I will. I hope this project finished this year.

This will be my last LGTM 🚀 A hug to all the beautiful members of NIAEFEUP.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
```diff
@@ -29,6 +37,25 @@ class AuthController(val authService: AuthService) {
...
+    }
+
+    @PostMapping("/password/recovery/{token}/confirm")
+    fun confirmRecoveryToken(
+        @Valid @RequestBody
+        dto: PasswordRecoveryConfirmDto,
+        @PathVariable token: String
+    ): Account {
+        authService.confirmRecoveryToken(token)
+        println("test")
```
MRita443:
```suggestion
```

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
```diff
@@ -141,6 +144,22 @@ class ErrorController(private val objectMapper: ObjectMapper) : ErrorController,
         return wrapSimpleError(e.message ?: "invalid authentication")
     }
 
+    @ExceptionHandler(JwtValidationException::class)
+    @ResponseStatus(HttpStatus.UNAUTHORIZED)
+    fun invalidAuthentication(e: JwtValidationException): CustomError {
+        return if (e.message?.contains("expired") == true) {
```
MRita443, line 150:
```suggestion
        return if (e.message?.contains("expired")) {
```

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AuthController.kt
MRita443: @rubuy-74 I think these lines need an explaining comment

MRita443 (changes requested):
Looks great, some small details and conversations I unresolved. Also when attempting to change password and giving a password too short, it returns the attempted password back to the user, maybe that's not a good idea

src/main/kotlin/pt/up/fe/ni/website/backend/service/AuthService.kt
rubuy-74: This is already resolved. I separated the updating password section to a different function that is called if the confirmRecoryToken function succeds

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
rubuy-74: This can't be because e.message?.contains("expired") return a Boolean? and the expression must return a Boolean. The "== true" is basically forcing it to be a valid Boolean and not null

MRita443 (approved):
LGTM

PR #115, opened 2023-02-21
Use global editorconfig and reformat files

### LINKED ISSUE #101: lint: Create configuration file
Currently, we're using the default configuration of ktlint in Github's action and most of the team members are also using the ktlint plugin in IntelliJ. However, the default of the plugin is different from what is running in the actions.

We should have a consistent configuration in the repo that's used by the plugin and in the actions.

bdmendes:
Closes #101 

Adds an [`.editorconfig`](https://editorconfig.org/) to the project to enforce standards such as max line length, spaces instead of tabs but also to provide kotlin style configuration proprietary to `ktlint`: plugin version, disabled rules and used style. The chosen style is [Android's](https://developer.android.com/kotlin/style-guide) instead of the official one, since it is more rigid and opinionated, which in my opinion is always a good thing.

# Review checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,4 +12,4 @@ max_line_length = 120
 [*.{kt,kts}]
 ktlint_code_style = android
 ktlint_version = 0.48.2
-ktlint_disabled_rules = import-ordering # ktlint IntelliJ IDEA extension seems to disagree with ktlint CLI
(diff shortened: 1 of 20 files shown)


ttoino (approved):
LGTM!

BrunoRosendo (approved):
Looks great!

PR #118, opened 2023-02-22
Added Logging interface

### LINKED ISSUE #32: global: add proper logging
Currently, the only logs we have are errors printed to `stderr`. We should have a better and consistent way of logging (e.g. error, info, debug, etc.). We can look at the available libraries and other Kotlin/Spring projects to choose.

MRita443:
Closes #32 

Adds a proper logging interface based on SLF4J API and Logback. Classes that require using the logger should now implement the Logging interface and refer to logger().

# Review checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,11 +29,14 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
     implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+    implementation("ch.qos.logback:logback-core:1.4.5")
+    implementation("org.slf4j:slf4j-api:2.0.6")
     annotationProcessor("org.springframework.boot:spring-boot-configuration-processor")
     implementation("org.springframework.boot:spring-boot-starter-validation:3.0.0")
     developmentOnly("org.springframework.boot:spring-boot-devtools")
     runtimeOnly("com.h2database:h2")
     testImplementation("org.springframework.boot:spring-boot-starter-test")
+    testImplementation("ch.qos.logback:logback-classic:1.4.5")
 }
 
 tasks.withType<KotlinCompile> {
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/config/Logger.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/config/Logger.kt
@@ -1,0 +1,8 @@
+package pt.up.fe.ni.website.backend.config
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory.getLogger
+
+interface Logging
+
+inline fun <reified T : Logging> T.logger(): Logger = getLogger(T::class.java)
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/config/Logging.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/config/Logging.kt
@@ -3,6 +3,6 @@ package pt.up.fe.ni.website.backend.config
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory.getLogger
 
-interface Logging
-
-inline fun <reified T : Logging> T.logger(): Logger = getLogger(T::class.java)
+interface Logging {
+    val logger: Logger get() = getLogger(this::class.java)
+}
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController.kt
@@ -15,7 +15,6 @@ import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.bind.annotation.RestControllerAdvice
 import pt.up.fe.ni.website.backend.config.Logging
-import pt.up.fe.ni.website.backend.config.logger
 
 data class SimpleError(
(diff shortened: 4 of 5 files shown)


BrunoRosendo (approved):
Looks good, simple and clean!

ttoino (approved):
LGTM!

bdmendes (commented):
Looks good! Have you tried the annotation as in https://stackabuse.com/guide-to-logging-in-spring-boot/? What do you think of that vs the interface @BrunoRosendo?

LuisDuarte1 (commented):
I agree with @bdmendes because an annotation looks cleaner, in my opinion, and it doesn't use inheritance, even though, `@Slf4j` is not a explicit name and it should be changed to something more straightforward.

BrunoRosendo (changes requested):
> Looks good! Have you tried the annotation as in https://stackabuse.com/guide-to-logging-in-spring-boot/? What do you think of that vs the interface @BrunoRosendo?

I don't have a strong opinion between inheritance and annotation but I'd prefer only having to do 1 import. Right now, we import Logging and then we also import `logger()` because it's an extension function. I think either the annotation or having `logger()` as a field of an abstract class (for example, there might be other ways) would be easier to use

(requested changes for discussion)

MRita443:
I tried the annotation method but I could not get it to work, as recommended [here](https://stackoverflow.com/questions/60419699/implement-slf4j-annotation-from-lombok-in-kotlin) I then tried looking into other idiomatic solutions in Kotlin for better alternatives (more detailed [explanation](https://stackoverflow.com/questions/34416869/idiomatic-way-of-logging-in-kotlin/65195488#65195488)). Here are a few that I liked:

- **Extension function without marking interface:** I initially used a marking interface because the standalone function would lead to the pollution of the any namespace, but that then led to having two imports, we could always remove the marking interface if the pollution would not be of significance;
- **Property Delegates:** A clean solution that would lead to a single import and a line of code of the sorts of `private val logger by LoggerDelegate()` on every class requiring logging;
- **Abstract class:** Another play on my first extension function with marker solution would be to create an abstract class that implements the interface, create a logger variable there, and import just the abstract class instead. A big set back of this would be that the logging messages refer to the abstract class instead of the class where the logger is being called.

Let me know what you guys think and if you have any other suggestions!

bdmendes:
> I tried the annotation method but I could not get it to work, as recommended [here](https://stackoverflow.com/questions/60419699/implement-slf4j-annotation-from-lombok-in-kotlin) I then tried looking into other idiomatic solutions in Kotlin for better alternatives (more detailed [explanation](https://stackoverflow.com/questions/34416869/idiomatic-way-of-logging-in-kotlin/65195488#65195488)). Here are a few that I liked:
> 
>     * **Extension function without marking interface:** I initially used a marking interface because the standalone function would lead to the pollution of the any namespace, but that then led to having two imports, we could always remove the marking interface if the pollution would not be of significance;
> 
>     * **Property Delegates:** A clean solution that would lead to a single import and a line of code of the sorts of `private val logger by LoggerDelegate()` on every class requiring logging;
> 
>     * **Abstract class:** Another play on my first extension function with marker solution would be to create an abstract class that implements the interface, create a logger variable there, and import just the abstract class instead. A big set back of this would be that the logging messages refer to the abstract class instead of the class where the logger is being called.
> 
> 
> Let me know what you guys think and if you have any other suggestions!

Personally, I think that your current solution is cleaner than all of those.
Just a quick suggestion: if I recall correctly, interfaces in Kotlin can have variables as well, so you can save an internal logger on the interface and refer to it every time `logger()` is called instead of calling `getLogger()`.

ttoino:
> > I tried the annotation method but I could not get it to work, as recommended [here](https://stackoverflow.com/questions/60419699/implement-slf4j-annotation-from-lombok-in-kotlin) I then tried looking into other idiomatic solutions in Kotlin for better alternatives (more detailed [explanation](https://stackoverflow.com/questions/34416869/idiomatic-way-of-logging-in-kotlin/65195488#65195488)). Here are a few that I liked:
> > ```
> > * **Extension function without marking interface:** I initially used a marking interface because the standalone function would lead to the pollution of the any namespace, but that then led to having two imports, we could always remove the marking interface if the pollution would not be of significance;
> > 
> > * **Property Delegates:** A clean solution that would lead to a single import and a line of code of the sorts of `private val logger by LoggerDelegate()` on every class requiring logging;
> > 
> > * **Abstract class:** Another play on my first extension function with marker solution would be to create an abstract class that implements the interface, create a logger variable there, and import just the abstract class instead. A big set back of this would be that the logging messages refer to the abstract class instead of the class where the logger is being called.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let me know what you guys think and if you have any other suggestions!
> 
> Personally, I think that your current solution is cleaner than all of those. Just a quick suggestion: if I recall correctly, interfaces in Kotlin can have variables as well, so you can save an internal logger on the interface and refer to it every time `logger()` is called instead of calling `getLogger()`.

You could do something like this with only one import:

```kotlin
interface Logging {
    val logger: Logger get() = getLogger(this::class.java)
}
```

Unfortunately you can't assign default values to properties in interfaces, as they are abstract and must be given a value by the implementing class. So something like this **would not** work:

```kotlin
interface Logging {
    val logger = getLogger(this::class.java)
}
```

bdmendes:
In that case, I think the current solution is fine. Let me know what you all think

BrunoRosendo (approved):
I'm OK with this implementation too

ttoino (approved):

Sirze01 (approved):

PR #119, opened 2023-02-22
Feature/password-change

### LINKED ISSUE #85: accounts: change password

coutinho21:
Closes #85 

The objective is to add a new endpoint to change an account's password.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/ChangePassDto.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/ChangePassDto.kt
@@ -1,0 +1,6 @@
+package pt.up.fe.ni.website.backend.dto.auth
+
+data class ChangePassDto(
+    val oldPassword: String,
+    val newPassword: String
+)
--- a/src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
+++ b/src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
@@ -10,6 +10,7 @@ import org.junit.jupiter.api.DisplayName
 import org.junit.jupiter.api.Test
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.MediaType
+import org.springframework.security.crypto.password.PasswordEncoder
 import org.springframework.test.web.servlet.MockMvc
 import org.springframework.test.web.servlet.get
 import org.springframework.test.web.servlet.post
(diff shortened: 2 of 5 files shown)


src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -420,6 +422,45 @@ class AccountControllerTest @Autowired constructor(
...
+    inner class ChangePassword {
+        private val password = "test_password"
+        private val changePasswordAccount: Account = ObjectMapper().readValue(
+            ObjectMapper().writeValueAsString(testAccount),
+            Account::class.java
+        )
+
+        @BeforeEach
+        fun addAccount() {
+            changePasswordAccount.password = encoder.encode(changePasswordAccount.password)
```
bdmendes: Can this line be in a before all or init-like method? Only needs to be done once (and would actually fail if there were more test methods inside this inner class)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/AccountControllerTest.kt
```diff
@@ -420,6 +422,45 @@ class AccountControllerTest @Autowired constructor(
...
+            repository.save(changePasswordAccount)
+        }
+
+        @Test
+        fun `should change password`() {
```
bdmendes, line 444: You should also test the sad path, ie. the provided password is wrong and the request should fail.

bdmendes (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/dto/auth/ChangePassDto.kt
```diff
@@ -0,0 +1,6 @@
+package pt.up.fe.ni.website.backend.dto.auth
+
+data class ChangePassDto(
```
ttoino: I think we should be consistent with our naming and always write out password in full

.github/workflows/ci.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         with:
           name: lint-reports
           path: build/reports/ktlint
+        if: always()
```
bdmendes: Is this here by mistake?

bdmendes (changes requested):

.github/workflows/ci.yml
coutinho21: Yes I had to clone the project again and create the branch so when i pulled everything that i had from this pr, it added @ttoino's commits that i had locally and this branch didn't have remotely.

coutinho21: Is it a problem?

bdmendes: That PR was already merged, so this should not be here. Can you fix that?

bdmendes (approved):

ttoino (approved):

PR #120, opened 2023-02-22
Fix ci artifacts not being uploaded when a job fails

### LINKED ISSUE #116: actions: test artefacts not being generated when tests fail

ttoino:
Closes #116.

# Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - master
       - develop
   pull_request:
-    branches: 
+    branches:
       - master
       - develop
 
@@ -42,6 +42,7 @@ jobs:
         with:
           name: lint-reports
           path: build/reports/ktlint
+        if: always()
 
   test:
     name: 'Test'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         with:
           name: lint-reports
           path: build/reports/ktlint
+        if: ${{ always() }}
```
limwa: Expressions in an if-statement don't require the `${{ }}` syntax
```suggestion
        if: always()
```

.github/workflows/ci.yml
```diff
@@ -75,3 +76,4 @@ jobs:
           path: |
             build/reports/tests/test
             build/reports/jacoco/test
+        if: ${{ always() }}
```
limwa: Same here!
```suggestion
        if: always()
```

.github/workflows/ci.yml
limwa: I don't think this is that big of a deal, I thought I would just let you know!

limwa (approved):
LGTM 🚀

BrunoRosendo (approved):

PR #139, opened 2023-03-20
64 projects hall of fame

### LINKED ISSUE #64: projects: hall of fame
Similarly to what's being done in #41, we should have another list of members representing a hall of fame. The idea behind this is to keep track of alumni that contributed to the project and have now left the organization.

We should have endpoints to add and remove members to/from the hall of fame. We also need endpoints to switch between hall of fame and active team

AvilaAndre:
Closes #64

Created a structure for the hall of fame.

Adds an account to the hall of fame, removes and account from the hall of fame and moves an account from the hall of fame to the team and vice-versa.

# Review checklist


src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
```diff
@@ -30,6 +30,10 @@ abstract class Activity(
     @OneToMany(fetch = FetchType.EAGER)
     open val teamMembers: MutableList<Account>,
 
+    @JoinColumn // Fixme: Don't know if this is needed
```
AvilaAndre: I don't know if this is needed, added it because it was on the structure above.

src/main/kotlin/pt/up/fe/ni/website/backend/service/ActivityService.kt
```diff
@@ -32,4 +36,59 @@ abstract class ActivityService<T : Activity>(
         activity.teamMembers.removeIf { it.id == idAccount }
         return repository.save(activity)
     }
+
+    fun addHallOfFameMemberById(idActivity: Long, idAccount: Long): T {
+        val activity = getActivityById(idActivity)
+        val account = accountService.getAccountById(idAccount)
+        if (activity.teamMembers.find { it.id == idAccount } != null) {
+            // TODO: deal with member that already exists on the team, should it return an error or move?
```
AvilaAndre: Should it throw an error or move the account when adding to the hall of fame while existing on the team.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -53,4 +53,28 @@ class ProjectController(private val service: ProjectService) {
...
+    fun moveMemberToActiveTeamById(
+        @PathVariable idProject: Long,
+        @PathVariable idAccount: Long
+    ) = service.moveMemberToActiveTeamById(idProject, idAccount)
+
+    @PutMapping("/{idProject}/moveMemberToHallOfFameById/{idAccount}")
+    fun moveMemberToHallOfFameById(
+        @PathVariable idProject: Long,
+        @PathVariable idAccount: Long
+    ) = service.moveMemberToHallOfFameById(idProject, idAccount)
```
BrunoRosendo: This seems a bit redundant imo. I'm trying to find use cases for these endpoints, maybe having an option in the UI to directly move members between team/hof? I think we should discuss this with more people, don't remove them for now

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
BrunoRosendo: This is needed in order to properly define the foreign key of the relationship (see https://www.baeldung.com/spring-data-rest-relationships for example)

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
```diff
@@ -30,6 +30,10 @@ abstract class Activity(
...
+    @OneToMany(fetch = FetchType.EAGER)
+    open val hallOfFame: MutableList<Account>,
```
BrunoRosendo: This shouldn't be in the activity, since the hall of fame is a feature for projects only (correct if I'm wrong)

src/main/kotlin/pt/up/fe/ni/website/backend/service/ActivityService.kt
```diff
@@ -32,4 +36,59 @@ abstract class ActivityService<T : Activity>(
         activity.teamMembers.removeIf { it.id == idAccount }
         return repository.save(activity)
     }
+
+    fun addHallOfFameMemberById(idActivity: Long, idAccount: Long): T {
```
BrunoRosendo: for the same reason as above, all of the logic here would be ported into ProjectService

src/main/kotlin/pt/up/fe/ni/website/backend/service/ActivityService.kt
```diff
@@ -16,6 +16,10 @@ abstract class ActivityService<T : Activity>(
     fun addTeamMemberById(idActivity: Long, idAccount: Long): T {
         val activity = getActivityById(idActivity)
         val account = accountService.getAccountById(idAccount)
+        if (activity.hallOfFame.find { it.id == idAccount } != null) {
+            // TODO: deal with member that already exists on the hallOfFame, should it return an error or move?
+            return moveMemberToActiveTeamById(idActivity, idAccount)
```
BrunoRosendo: I think the logic makes sense but it **needs** to be documented in the API, since it's not straight forward

src/main/kotlin/pt/up/fe/ni/website/backend/service/ActivityService.kt
```diff
@@ -32,4 +36,59 @@ abstract class ActivityService<T : Activity>(
...
+                ErrorMessages.accountNotFound(
+                    idAccount
+                )
+            )
+        }
+        val account = accountService.getAccountById(idAccount)
+        if (activity.teamMembers.removeIf { it.id == idAccount }) {
+            activity.hallOfFame.add(account)
+        }
+        return repository.save(activity)
```
BrunoRosendo: If we keep this, I think we can reuse functions like `removeHallOfFameMemberById`, `removeTeamMemberById`, and `add` versions to avoid repeating code

BrunoRosendo (changes requested):
Good job so far! Besides my comments, I'd like to have more opinions on this implementation because a lot of the code is identical to the team members' implementation, so I'm wondering if there's a better way to do this.

Other than that, the PR is also missing tests/docs (see wiki) but I assume it's because it is still in draft

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
bdmendes: I think the same

src/main/kotlin/pt/up/fe/ni/website/backend/service/ActivityService.kt
bdmendes: I think that this should not be mandatory

BrunoRosendo:
@AvilaAndre is this ready for review?

AvilaAndre:
@BrunoRosendo I changed the payloads, but I still need to create tests for my own changes.

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/ProjectService.kt
```diff
@@ -74,4 +79,24 @@ class ProjectService(
...
+    fun removeHallOfFameMemberById(idProject: Long, idAccount: Long): Project {
+        val project = getProjectById(idProject)
+        if (!accountService.doesAccountExist(idAccount)) {
+            throw NoSuchElementException(
+                ErrorMessages.accountNotFound(
+                    idAccount
+                )
+            )
+        }
+        project.hallOfFame.removeIf { it.id == idAccount }
```
AvilaAndre, line 99: This is the line the code coverage bot warns about, I guess changing its behavior would fix it, but it would do the same thing anyway. I think there is no need to update it.

BrunoRosendo: solved by changing codecov config

src/test/kotlin/pt/up/fe/ni/website/backend/controller/ProjectControllerTest.kt
```diff
@@ -58,9 +58,24 @@ internal class ProjectControllerTest @Autowired constructor(
         )
     )
 
+    final val testOldAccount = Account(
```
BrunoRosendo: any reason to call it "old"? Seems a bit odd, I think something like testAccount2 would be easier to understand

src/test/kotlin/pt/up/fe/ni/website/backend/controller/ProjectControllerTest.kt
```diff
@@ -868,6 +898,140 @@ internal class ProjectControllerTest @Autowired constructor(
...
+            )
+                .andExpectAll(
+                    status().isOk,
+                    content().contentType(MediaType.APPLICATION_JSON),
+                    jsonPath("$.hallOfFame.length()").value(0)
+                )
+                .andDocument(
+                    documentation,
+                    "Remove account from Project's Hall of Fame",
+                    "This operation adds an account to a given project's hall of fame.",
```
BrunoRosendo:
```suggestion
                    "This operation removes an account from a given project's hall of fame.",
```

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadActivity.kt
```diff
@@ -13,6 +13,11 @@ class PayloadActivity {
                 DocumentedJSONField("id", "Id of the activity", JsonFieldType.NUMBER),
                 DocumentedJSONField("title", "Title of the activity", JsonFieldType.STRING),
                 DocumentedJSONField("description", "Description of the activity", JsonFieldType.STRING),
+                DocumentedJSONField(
+                    "hallOfFame",
+                    "Array of members that were once associated with the project",
+                    JsonFieldType.ARRAY
+                ),
```
BrunoRosendo, line 20:
```suggestion
```
not part of activity

BrunoRosendo (changes requested):
Good job, it's almost done! I just left some minor comments

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadActivity.kt
BrunoRosendo: This was because Spring includes project/event's properties when serializing activities. This is unexpected but shouldn't be a problem

BrunoRosendo (approved):
Good job! It'd be cool if you improve the commit history, otherwise all is well

src/test/kotlin/pt/up/fe/ni/website/backend/controller/ProjectControllerTest.kt
```diff
@@ -241,6 +262,7 @@ internal class ProjectControllerTest @Autowired constructor(
                     content().contentType(MediaType.APPLICATION_JSON),
                     jsonPath("$.title").value(testProject.title),
                     jsonPath("$.description").value(testProject.description),
+                    jsonPath("$.hallOfFame.length()").value(1),
```
MRita443, line 265: I think here you could not only verify the length of the list but also that the correct element was added (ex. by checking the email)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/ProjectControllerTest.kt
```diff
@@ -868,6 +898,140 @@ internal class ProjectControllerTest @Autowired constructor(
...
+        fun addToRepositories() {
+            accountRepository.save(testAccount)
+            accountRepository.save(testAccount2)
+            accountRepository.save(newAccount)
+            repository.save(testProject)
+        }
+
+        private val parameters = listOf(
+            parameterWithName("idProject").description(
+                "ID of the project to add the account to the hall of fame"
```
MRita443: Something along these lines would be clearer imo
```suggestion
                "ID of the project whose hall of fame the account will be added to"
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/ProjectControllerTest.kt
```diff
@@ -868,6 +898,140 @@ internal class ProjectControllerTest @Autowired constructor(
...
+        @BeforeEach
+        fun addToRepositories() {
+            accountRepository.save(testAccount)
+            accountRepository.save(testAccount2)
+            repository.save(testProject)
+        }
+
+        private val parameters = listOf(
+            parameterWithName("idProject").description(
+                "ID of the project to remove the account from the hall of fame"
```
MRita443:
```suggestion
                "ID of the project whose hall of fame the account will be removed from"
```

MRita443 (changes requested):
Good job! Really minor suggestions

MRita443 (approved):
Great job! :)

Issue #143, opened 2023-03-28
Refactor endpoints to follow a more "REST-like" approach


BrunoRosendo:
Our API currently work fine but there are some endpoint semantics that don't make sense when thinking about the REST standards.
The backend will mostly be an internal tool but it would be great to follow the standards and re-think the API before it starts being heavily used by the frontend (or possibly other services).

You can find a lot of information online about REST best practices and recommendations, but here's one I found useful: https://www.freecodecamp.org/news/rest-api-best-practices-rest-endpoint-design-examples/

I'll also leave some suggestions of recommendations we're not currently following and the respective endpoints.

### Use Nouns Instead of Verbs in Endpoints

This is one of the biggest flaws in our API that was pointed out in one of our meetings. Here are some examples where we violated this (please ensure no more were added in the meantime):
- `POST /accounts/changePassword/{id}` This could probably be changed to something like `POST /accounts/{id}/password`
- `PUT /events/{eventId}/addTeamMember/{accountId}` to `PUT /events/{eventId}/team/{accountId}` or `POST /events/{eventId}/members` with the accountId in the body. Some discussion of a similar problem [here](https://stackoverflow.com/questions/6324547/how-can-i-handle-many-to-many-relationships-in-a-restful-api)
- Same thing as above for removing team members and for projects too. Same for hall of fame (in #139 )
- `PUT /projects/{id}/archive` I'm not sure about this one, I like the endpoint but it violates the "rule". Same for unarchive.

### Redundant words in endpoint path

For example, if we have a POST request, there's no need to include a word like `new` in the path. This is redundant and just complicates things:

- `POST /accounts/new` to `POST /accounts` (and similar ones for creating entities)
- `POST /auth/new` to `POST /auth` maybe

### Use Filtering, Sorting, and Pagination to Retrieve the Data Requested

- Remove `GET /events/category/{category}` and add a query filter to `GET /events` with the same functionality, like `GET /events?category={category}`

### Ambiguous endpoints

It might be a problem to have endpoint paths not clearly explaining what they do:

- `GET /generations` to  `GET /generations/years`

Please give your thoughts about the suggestions and suggest new changes. Be sure to make your own investigations instead of following my suggestions blindly.

PR #145, opened 2023-04-05
Role management endpoints

### LINKED ISSUE #48: projects: add roles across team members
In order to provide authorization for editing/deleting projects to only some of its members, we should implement a system of permissions, similar to what's being done in #3 and #12 

A simple approach would be to simply have a project coordinator leader defined but that wouldn't support multiple leaders and doesn't seem very scalable. It would probably be best to use the role functionality to identify the project leaders or other roles

### LINKED ISSUE #79: roles: Create endpoints for role management
After the work done in #49, where the roles and permissions were created and associated with accounts and activities, we now need to use them to implement our role management system.

Note that managing access from permissions is not part of this issue, which should be done in or after #60. That said, this issue should tackle (more or less) the following tasks:

- [ ] Create controllers for the following features (edit to include more if needed):
  - [ ] Create and delete roles associated with a set of permissions. Watch out for the activity permissions too (e.g. `PerActivityRole` model and `getEffectivePermissionsForActivity` function in `Account` model)
  - [ ] Add/Revoke permissions to/from roles, including activity roles. One can also consider a similar implementation to the "create" endpoint, identical to other controllers where entities are "recreated".
  - [ ] Add/Revoke roles to/from accounts. "Recreating" an account is not a good idea here
- [ ] Make sure to create integration tests (and unit if necessary) for the endpoints you create. Since #49 didn't implement any controllers, only unit tests are present, meaning that testing the roles' functionalities is a very important part of this issue.
- [ ] Watch out for the server's responses. There might be cases where role information is unexpectedly returned and cases where something important is missing. This is due to the choice between lazy and eager loading when serializing the models (check discussion in https://github.com/NIAEFEUP/website-niaefeup-backend/pull/49#issuecomment-1376139238). Update fetch type when necessary

LuisDuarte1:
Closes #79 

Adds various endpoints for role management, such as:
 - Add/remove roles
 - Gets all roles
 - Get a specific role
 - Add/remove users from a role
 - Grant/Revoke permissions from a role
 - Add/remove activities that are associated with a role
 - Add/remove permissions on PerActivityRoles


BrunoRosendo:
Is it ready for review?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+@RestController
+@RequestMapping("/roles")
+class RoleController(private val roleService: RoleService) {
+    @GetMapping
+    fun getAllRoles() = roleService.getAllRoles()
+
+    @GetMapping("/{id}")
+    fun getRole(@PathVariable id: Long) = roleService.getRole(id)
+
+    @PostMapping("/new")
```
BrunoRosendo:
```suggestion
    @PostMapping
```
`/new` is redundant, as mentioned in #143

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun createNewRole(@RequestBody dto: RoleDto) = roleService.createNewRole(dto)
+
+    @DeleteMapping("/{id}")
+    fun deleteRole(@PathVariable id: Long): Map<String, String> {
+        roleService.deleteRole(id)
+        return emptyMap()
+    }
+
+    @PostMapping("/{id}/grant")
+    fun grantPermissionRole(
```
BrunoRosendo:
```suggestion
    fun grantPermissionToRole(
```

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+        @PathVariable id: Long,
+        @RequestBody permissionsDto: PermissionsDto
+    ): Map<String, String> {
+        roleService.grantPermissionToRole(id, Permissions(permissionsDto.permissions))
```
BrunoRosendo: Is there any reason why you can't serialize `Permissions` directly? I don't think #34 applies here since Permissions isn't an entity

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+        return emptyMap()
+    }
+
+    @PostMapping("/{id}/revoke")
+    fun revokePermissionRole(
```
BrunoRosendo:
```suggestion
    fun revokePermissionFromRole(
```
just to be consistent

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+    @PostMapping("/new")
+    fun createNewRole(@RequestBody dto: RoleDto) = roleService.createNewRole(dto)
+
+    @DeleteMapping("/{id}")
+    fun deleteRole(@PathVariable id: Long): Map<String, String> {
+        roleService.deleteRole(id)
+        return emptyMap()
+    }
+
+    @PostMapping("/{id}/grant")
```
BrunoRosendo: What do you think about using `/{id}/permissions` and use POST/DELETE to differentiate between granting and revoking permissions? It's more according to REST standards

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun revokePermissionRole(
+        @PathVariable id: Long,
+        @RequestBody permissionsDto: PermissionsDto
+    ): Map<String, String> {
+        roleService.revokePermissionFromRole(id, Permissions(permissionsDto.permissions))
+        return emptyMap()
+    }
+
+    @PostMapping("/{id}/users")
+    fun addUserToRole(@PathVariable id: Long, @RequestBody userId: Long): Map<String, String> {
```
BrunoRosendo: Do you pass the userId directly in the body? I'm more fan of something like `{ userId: 2 }`, wdyt?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+        return emptyMap()
+    }
+
+    @DeleteMapping("/{id}/users")
+    fun removeUserFromRole(@PathVariable id: Long, @RequestBody userId: Long): Map<String, String> {
+        roleService.removeUserFromRole(id, userId)
+        return emptyMap()
+    }
+
+    @PostMapping("/{id}/activity/{activityId}/permissions")
```
BrunoRosendo:
```suggestion
    @PostMapping("/{id}/activities/{activityId}/permissions")
```
Similarly to previous cases

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+    ): Map<String, String> {
+        roleService.grantPermissionToRoleOnActivity(
+            id,
+            activityId,
+            Permissions(permissionsDto.permissions)
+        )
+        return emptyMap()
+    }
+
+    @DeleteMapping("/{id}/activity/{activityId}/permissions")
```
BrunoRosendo: same as above
```suggestion
    @DeleteMapping("/{id}/activities/{activityId}/permissions")
```

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun revokePermissionToPerActivityRole(
```
BrunoRosendo:
```suggestion
    fun revokePermissionFromPerActivityRole(
```

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
 
 @Service
+@Transactional
 class RoleService(
     private val roleRepository: RoleRepository,
-    private val perActivityRoleRepository: PerActivityRoleRepository
+    private val perActivityRoleRepository: PerActivityRoleRepository,
+    private val generationRepository: GenerationRepository,
+    private val accountRepository: AccountRepository,
+    private val activityRepository: ActivityRepository<Activity>
```
BrunoRosendo: Small detail but it's often better to use the services instead of repositories of other entities, since you already have access to better abstractions related to our system.

Of course, this is only the case if there are no circular dependencies so please check for that (I didn't do it myself), not all situations apply.

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
 ) {
 
-    fun grantPermissionToRole(role: Role, permission: Permission) {
-        role.permissions.add(permission)
+    fun getRole(roleId: Long): Role {
```
BrunoRosendo:
```suggestion
    fun getRoleById(roleId: Long): Role {
```
Small detail, just to be according to other services. Same applies to other functions below

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
+        val role = roleRepository.findById(roleId).orElseThrow {
+            throw NoSuchElementException(ErrorMessages.roleNotFound(roleId))
+        }
+        return role
+    }
+    fun getAllRoles(): List<Role> = roleRepository.findAll().toList()
+    fun grantPermissionToRole(roleId: Long, permissions: Permissions) {
+        val role = roleRepository.findById(roleId).orElseThrow {
+            throw NoSuchElementException(ErrorMessages.roleNotFound(roleId))
+        }
```
BrunoRosendo: you can use `getRole()` to avoid code duplication

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
+        role.permissions.addAll(permissions)
         roleRepository.save(role)
     }
 
-    fun revokePermissionFromRole(role: Role, permission: Permission) {
-        role.permissions.remove(permission)
+    fun revokePermissionFromRole(roleId: Long, permissions: Permissions) {
+        val role = roleRepository.findById(roleId).orElseThrow {
+            throw NoSuchElementException(ErrorMessages.roleNotFound(roleId))
+        }
```
BrunoRosendo: same as above

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
     }
 
-    fun grantPermissionToRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
+    fun grantPermissionToRoleOnActivity(roleId: Long, activityId: Long, permissions: Permissions) {
+        val activity = activityRepository.findById(activityId).orElseThrow {
+            throw NoSuchElementException(ErrorMessages.activityNotFound(activityId))
+        }
+        val role = roleRepository.findById(roleId).orElseThrow {
+            throw NoSuchElementException(ErrorMessages.roleNotFound(roleId))
+        }
```
BrunoRosendo: same as above for using services

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
         val foundActivity = activity.associatedRoles
```
BrunoRosendo:
```suggestion
        val foundPerActivityRole = activity.associatedRoles
```
previous name is confusing, maybe something like this?

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
         foundActivity.activity = activity
 
-        foundActivity.role = role
-        foundActivity.permissions.add(permission)
+        foundActivity.permissions.addAll(permissions)
         perActivityRoleRepository.save(foundActivity)
+        if (activity.associatedRoles.find { it.activity == activity } == null) {
+            role.associatedActivities.add(foundActivity)
+        }
+        activityRepository.save(activity)
```
BrunoRosendo: Is it necessary to save the activity role in both repositories? In my experience, it's either for testing reasons (which fair enough but please comment) or there's something wrong with the relationships.

I say it's ok for testing reasons because sometimes the entities are not updated while a test is running (transaction shenanigans)

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
         val foundActivity = activity.associatedRoles
             .find { it.role == role } ?: return
-
-        foundActivity.permissions.remove(permission)
+        foundActivity.permissions.removeAll(permissions)
         perActivityRoleRepository.save(foundActivity)
+        activityRepository.save(activity)
+    }
+
+    fun createNewRole(dto: RoleDto): Role {
```
BrunoRosendo: I think we should have an optional parameter to create roles for older generations. The same applies for other operations but code might not change in all of them

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,130 @@
...
+        return role
+    }
+
+    fun deleteRole(roleId: Long) {
+        val role = roleRepository.findByIdOrNull(roleId)
+            ?: throw NoSuchElementException(ErrorMessages.roleNotFound(roleId))
+        val latestGeneration = generationRepository.findFirstByOrderBySchoolYearDesc()!!
+        latestGeneration.roles.remove(role)
+        generationRepository.save(latestGeneration)
+        roleRepository.deleteById(roleId)
```
BrunoRosendo: Isn't the generation updated when the role is deleted? I thought it'd do that automatically 😢

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
```diff
@@ -0,0 +1,22 @@
...
+import pt.up.fe.ni.website.backend.utils.documentation.Tag
+import pt.up.fe.ni.website.backend.utils.documentation.utils.DocumentedJSONField
+import pt.up.fe.ni.website.backend.utils.documentation.utils.ModelDocumentation
+
+class PayloadRoles : ModelDocumentation(
+    Tag.ROLES.name.lowercase(),
+    Tag.ROLES,
+    mutableListOf(
+        DocumentedJSONField("name", "Name of the role", JsonFieldType.STRING),
+        DocumentedJSONField("permissions", "Permissions in the role, as an integer", JsonFieldType.ARRAY),
```
BrunoRosendo, line 13: Is there a way we could list the permissions and respective integers? I might be asking for too much but @Sirze01 cc

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
```diff
@@ -0,0 +1,22 @@
...
+        DocumentedJSONField("isSection", "Role is section", JsonFieldType.BOOLEAN),
```
BrunoRosendo:
```suggestion
        DocumentedJSONField("isSection", "Whether the role should be displayed as a section", JsonFieldType.BOOLEAN),
```

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
```diff
@@ -0,0 +1,22 @@
...
+        DocumentedJSONField("id", "Internal ID of role", JsonFieldType.NUMBER),
+        DocumentedJSONField(
+            "associatedActivities",
+            "List of activities that are associated with this role",
```
BrunoRosendo: I think this is misleading, this is the list of roles associated with activities. The name is already confusing so the description should be more detailed

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+    )
+    val otherTestRole = Role(
+        "Coordenador de Projetos",
+        Permissions(listOf(Permission.EDIT_ACCOUNT)),
+        false
+    )
+
+    val testProject = Project(
+        "UNI",
+        "Melhor app"
```
BrunoRosendo: :no_mouth: 👀

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+        @Test
+        fun `should return testRole when provided by its id`() {
+            mockMvc.perform(get("/roles/${testRole.id}")).andExpectAll(
+                status().isOk,
+                content().contentType(MediaType.APPLICATION_JSON),
+                content().json(objectMapper.writeValueAsString(testRole))
+            ).andDocument(
+                documentationRoles,
+                "Returns a specific role",
+                "Returns a summary brief of a specific role, which makes getting one role way more efficient. "
```
BrunoRosendo:
```suggestion
                "Returns a summary brief of a specific role, which makes getting one role way more efficient."
```
😛

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            mockMvc.perform(get("/roles/4020")).andExpectAll(
+                status().isNotFound(),
+                content().contentType(MediaType.APPLICATION_JSON),
+                jsonPath("$.errors.length()").value(1),
+                jsonPath("$.errors.[0].message").value("role not found with id 4020")
+            ).andDocumentErrorResponse(documentationRoles, hasRequestPayload = true)
+        }
+    }
+
+    @NestedTest
```
BrunoRosendo, line 172: missing display name

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+    inner class CreateNewRole {
+
+        @BeforeEach
+        fun addRole() {
+            generationRepository.save(testGeneration)
+            roleRepository.save(testRole)
+            testGeneration.roles.add(testRole)
+            roleRepository.save(testRole)
+            generationRepository.save(testGeneration)
```
BrunoRosendo: care to explain? xD

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+        }
+
+        @AfterEach
+        fun removeRole() {
+            testGeneration.roles.remove(testRole)
+            testGeneration.roles.remove(otherTestRole)
+            roleRepository.save(testRole)
+            generationRepository.save(testGeneration)
+        }
```
BrunoRosendo: The db is cleared after each test so there shouldn't be a reason to do this

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                                "associatedActivities" to otherTestRole.associatedActivities,
+                                "accounts" to otherTestRole.accounts
+                            )
+                        )
+                    )
+            ).andExpectAll(
+                status().isOk,
+                content().contentType(MediaType.APPLICATION_JSON),
+                jsonPath("$.name").value(otherTestRole.name),
+                jsonPath("$.id").value(2), // this must be hardcoded
```
BrunoRosendo, line 205: you could just verify that it exists to prevent future mistakes. It doesn't really matter and the id rule could be something other than sequential

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                jsonPath("$.permissions.length()").value(otherTestRole.permissions.size),
+                jsonPath("$.isSection").value(otherTestRole.isSection),
+                jsonPath("$.associatedActivities").value(otherTestRole.associatedActivities)
+            ).andDocument(
+                documentationRoles,
+                "This creates a role, returning the complete role information",
+                "It will only create the role if it doesn't exist any other role with the same name"
```
BrunoRosendo:
```suggestion
                "It will only create the role if it no other role with the same name exists in that generation"
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            )
+            assert(roleRepository.existsById(2))
+            assert(generationRepository.findFirstByOrderBySchoolYearDesc() != null)
+            assert(generationRepository.findFirstByOrderBySchoolYearDesc()!!.roles.size == 2)
+            assert(generationRepository.findFirstByOrderBySchoolYearDesc()!!.roles[1].id!!.compareTo(2) == 0)
```
BrunoRosendo, line 217: why not a normal comparison?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                status().isNotFound(),
+                content().contentType(MediaType.APPLICATION_JSON),
+                jsonPath("$.errors.length()").value(1),
+                jsonPath("$.errors.[0].message").value("role not found with id 4020")
+            ).andDocumentErrorResponse(documentationRoles, hasRequestPayload = true)
+        }
+    }
+
+    @NestedTest
+    inner class CreateNewRole {
```
BrunoRosendo, line 174: you could test creating roles with accounts and perActivityRoles as well

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                jsonPath("$.errors.[0].message").value("role already exists")
+            )
+                .andDocumentErrorResponse(documentationRoles, hasRequestPayload = true)
+            assert(roleRepository.findByName(testRole.name) != null)
+            assert(generationRepository.findFirstByOrderBySchoolYearDesc() != null)
+            assert(generationRepository.findFirstByOrderBySchoolYearDesc()!!.roles.size == 2)
+        }
+    }
+
+    @NestedTest
```
BrunoRosendo, line 283: display name (more below)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                "Removes the role by its id",
+                "The id must exist in order to remove it correctly"
+            )
+            assert(roleRepository.findByName(testRole.name) == null)
+            assert(generationRepository.findFirstByOrderBySchoolYearDesc() != null)
+            assert(generationRepository.findFirstByOrderBySchoolYearDesc()!!.roles.size == 0)
+        }
+
+        @Test
+        fun `should not remove role id that does not exist`() {
```
BrunoRosendo:
```suggestion
        fun `should not remove role if id does not exist`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                content().contentType(MediaType.APPLICATION_JSON),
+                jsonPath("$.errors.length()").value(1),
+                jsonPath("$.errors.[0].message").value("role not found with id 1234")
+            )
+                .andDocumentErrorResponse(documentationRoles, hasRequestPayload = true)
+        }
+    }
+
+    @NestedTest
+    inner class GrantPermissionRole {
```
BrunoRosendo:
```suggestion
    inner class GrantPermissionToRole {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                .andDocumentEmptyObjectResponse(
+                    documentationPermissions,
+                    "Adds a set of permissions to a role by ID",
+                    "This doesn't check if the permission is already added, the role ID must be valid"
+                )
+            assert(roleRepository.findByIdOrNull(testRole.id!!)!!.permissions.contains(Permission.SUPERUSER))
+        }
+
+        @Test
+        fun `should grant permission to role that does not exists`() {
```
BrunoRosendo:
```suggestion
        fun `should not grant permission to role that does not exists`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+    @NestedTest
+    inner class GrantPermissionRole {
+
+        @BeforeEach
+        fun addRole() {
+            roleRepository.save(testRole)
+        }
+
+        @Test
+        fun `should grant permission to role that exists`() {
```
BrunoRosendo, line 348: can you test adding permissions to role with no permissions? The code is different in that case (code coverage in intellij helps with cases like this)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                        )
+                    )
+            ).andExpectAll(
+                status().isNotFound()
+            ).andDocumentErrorResponse(documentationPermissions, hasRequestPayload = false)
+        }
+    }
+
+    @NestedTest
+    inner class RevokePermissionRole {
```
BrunoRosendo:
```suggestion
    inner class RevokePermissionFromRole {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+
+        @BeforeEach
+        fun addRole() {
+            roleRepository.save(testRole)
+        }
+
+        @Test
+        fun `should revoke permission to role that exists`() {
```
BrunoRosendo:
```suggestion
        fun `should revoke permission from role that exists`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                            )
+                        )
+                    )
+            )
+                .andExpectAll(
+                    status().isOk
+                ).andDocumentEmptyObjectResponse(
+                    documentationPermissions,
+                    "Revokes permission by role ID",
+                    "Revokes permissions, it doesn't check if the permissions to be revoked are already off."
```
BrunoRosendo:
```suggestion
                    "Revokes permissions, it doesn't check if the role contains them."
```
seems simpler, wdyt?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                )
+            assert(!roleRepository.findByIdOrNull(testRole.id!!)!!.permissions.contains(Permission.SUPERUSER))
+        }
+
+        @Test
+        fun `should revoke permission to role that does not exists`() {
```
BrunoRosendo:
```suggestion
        fun `should not revoke permission from role that does not exist`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            mockMvc.perform(
+                post("/roles/${testRole.id}/users")
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .content(objectMapper.writeValueAsString(testAccount.id))
+            ).andExpectAll(
+                status().isOk
+            ).andDocumentEmptyObjectResponse(
+                documentationAccount,
+                "Add an account to a role by its IDs",
+                "It will return an error if the user is already in the role or if the role doesn't exist"
```
BrunoRosendo:
```suggestion
                "It will return an error if the user already has the role or if it doesn't exist"
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+        }
+
+        @Test
+        fun `should not add an non existing account to an existing role`() {
+            mockMvc.perform(
+                post("/roles/${testRole.id}/users")
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .content(objectMapper.writeValueAsString(1234))
+            ).andExpectAll(
+                status().isNotFound()
```
BrunoRosendo, line 492: maybe check for the error message in cases like this

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            roleRepository.save(testRole)
+            testRole.accounts.add(testAccount)
+            roleRepository.save(testRole)
+        }
+
+        @AfterEach
+        fun removeRolesAndUser() {
+            testRole.accounts.remove(testAccount)
+            roleRepository.save(testRole)
+        }
```
BrunoRosendo: same as above

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+
+        @Test
+        fun `should remove an existing account to an existing role`() {
```
BrunoRosendo:
```suggestion
        fun `should remove an existing account from an existing role`() {
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            mockMvc.perform(
+                delete("/roles/${testRole.id}/users")
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .content(objectMapper.writeValueAsString(testAccount.id))
+            ).andExpectAll(
+                status().isOk
+            ).andDocumentEmptyObjectResponse(
+                documentationAccount,
+                "Removes the account on the role by ID",
```
BrunoRosendo:
```suggestion
                "Removes the account from the role by ID",
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                "It only works if the user is in role and role exists"
```
BrunoRosendo:
```suggestion
                "It only works if the user has the role and it exists"
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            )
+            assert(roleRepository.findByIdOrNull(testRole.id!!)!!.accounts.size == 0)
+        }
+
+        @Test
+        fun `should not remove an non existing account to an existing role`() {
```
BrunoRosendo:
```suggestion
        fun `should not remove a non existing account from an existing role`() {
```

BrunoRosendo: Please check other cases with this grammar mistake

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                    .content(
+                        objectMapper.writeValueAsString(
+                            mapOf("permissions" to Permissions(listOf(Permission.EDIT_ACTIVITY)))
+                        )
+                    )
+            ).andExpectAll(
+                status().isOk
+            ).andDocumentEmptyObjectResponse(
+                documentationPermissions,
+                "Adds an permission to an role activity",
```
BrunoRosendo: a instead of an

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            testProject.associatedRoles.add(perActivityRole)
+            projectRepository.save(testProject)
+            roleRepository.save(testRole)
+        }
+
+        @AfterEach
+        fun removeAll() {
+            testProject.associatedRoles.removeAt(0)
+            projectRepository.save(testProject)
+        }
```
BrunoRosendo: same as above

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+                        )
+                    )
+
+            ).andExpectAll(
+                status().isOk
+            ).andDocumentEmptyObjectResponse(
+                documentationPermissions,
+                "Removes an PerRoleActivity permission by ID",
+                "It will not create a PerRoleActivity if it doesn't exist, it will simply return, " +
+                    "doesn't check if the permissions already are revoked..."
```
BrunoRosendo:
```suggestion
                    "doesn't check if the permissions are already revoked..."
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/RoleControllerTest.kt
```diff
@@ -0,0 +1,720 @@
...
+            assert(projectRepository.findByIdOrNull(testProject.id!!)!!.associatedRoles.size == 1)
+            assert(
+                projectRepository.findByIdOrNull(testProject.id!!)!!.associatedRoles[0].permissions.contains(
+                    Permission.EDIT_ACTIVITY
+                )
+            )
+        }
+
+        @Test
+        fun `should not remove an existing role activity perm on a non existing activity and non existing role`() {
```
BrunoRosendo: I'm confused, what's the meaning of this test?

BrunoRosendo (changes requested):
Good job so far! I know I wrote a lot of comments but the functionalities are there and seem to work as expected! A lot of the comments are about small things like documentation and code suggestions

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
LuisDuarte1: By serializing Permissions directly, the request body becomes `[1,2,3]`, for example, and with the DTO it becomes `{permissions:[1,2,3]}`which in my opinion, is more readable and more consistent with other API calls, wdyt? 

Maybe we need a solution for one parameter solutions like this, and userId, because creating an DTO for every one parameter body is not very sustainable.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
LuisDuarte1: Done

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
LuisDuarte1: I agree that we should do it for the sake of consistency in the API but creating an DTO for every situation like this and described above, seems not very sustainable, maybe I'm missing something

LuisDuarte1: But I'll add it anyways, until we find a better solution, or settle with creating a DTO for every one parameter body

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: Yeah it makes sense, this is fixed except for the generations which returns the dto instead of the generation itself, maybe that needs a small refactor in order to make services use other services instead of the repositories

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,91 @@
...
+    fun addUserToRole(@PathVariable id: Long, @RequestBody userIdDto: UserIdDto): Map<String, String> {
+        roleService.addUserToRole(id, userIdDto.userId)
+        return emptyMap()
+    }
+
+    @DeleteMapping("/{id}/users")
+    fun removeUserFromRole(@PathVariable id: Long, @RequestBody userIdDto: UserIdDto): Map<String, String> {
+        roleService.removeUserFromRole(id, userIdDto.userId)
+        return emptyMap()
+    }
```
bdmendes, line 63: I'm not sure if this is better here or in the accounts controller (which is really a detail)

LuisDuarte1: It is a detail, but I think it can work both ways, but for me at least makes more sense to add a user through the role

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
BrunoRosendo: I see and agree with you. I'm not aware of a better solution, I think DTOs are the standard go-to in cases like this one

src/main/kotlin/pt/up/fe/ni/website/backend/config/security/AuthConfig.kt
```diff
@@ -30,6 +30,7 @@ class AuthConfig(
 ) {
     @Bean
     fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
+        http.headers().frameOptions().disable()
```
bdmendes: Why is this needed?

LuisDuarte1: I forgot to remove it :smile: , it was to make the h2 console run locally because it uses a lot of iframes, I'll remove it

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,113 @@
...
+    fun revokePermissionFromRoleOnActivity(roleId: Long, activityId: Long, permissions: Permissions) {
+        val activity = activityService.getActivityById(activityId)
+        val role = getRole(roleId)
         val foundActivity = activity.associatedRoles
             .find { it.role == role } ?: return
-
-        foundActivity.permissions.remove(permission)
+        foundActivity.permissions.removeAll(permissions)
         perActivityRoleRepository.save(foundActivity)
+        // activityRepository.save(activity)
```
bdmendes: Comment

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,88 @@
...
+    @PostMapping("/{id}/activities/{activityId}/permissions")
+    fun addPermissionToPerActivityRole(
+        @PathVariable id: Long,
+        @PathVariable activityId: Long,
+        @RequestBody permissionsDto: PermissionsDto
+    ): Map<String, String> {
+        roleService.grantPermissionToRoleOnActivity(
+            id,
+            activityId,
+            Permissions(permissionsDto.permissions)
```
BrunoRosendo:
```suggestion
            permissionsDto.permissions
```
I think the constructor is redundant

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,88 @@
...
+    @DeleteMapping("/{id}/activities/{activityId}/permissions")
+    fun revokePermissionFromPerActivityRole(
+        @PathVariable id: Long,
+        @PathVariable activityId: Long,
+        @RequestBody permissionsDto: PermissionsDto
+    ): Map<String, String> {
+        roleService.revokePermissionFromRoleOnActivity(
+            id,
+            activityId,
+            Permissions(permissionsDto.permissions)
```
BrunoRosendo:
```suggestion
            permissionsDto.permissions
```

src/main/kotlin/pt/up/fe/ni/website/backend/dto/permissions/UserIdDto.kt
```diff
@@ -0,0 +1,5 @@
+package pt.up.fe.ni.website.backend.dto.permissions
+
+data class UserIdDto(
+    val userId: Long
+)
```
BrunoRosendo: I feel like this could be moved into `auth` and used for other requests when necessary, not just for permissions. I would personally rename `permissions` package to something like `roles' to be more broad

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
```diff
@@ -0,0 +1,88 @@
...
+@RequestMapping("/roles")
+class RoleController(private val roleService: RoleService) {
+    @GetMapping
+    fun getAllRoles() = roleService.getAllRoles()
+
+    @GetMapping("/{id}")
+    fun getRole(@PathVariable id: Long) = roleService.getRole(id)
+
+    @PostMapping
+    fun createNewRole(@RequestBody dto: RoleDto) = roleService.createNewRole(dto)
```
BrunoRosendo: We should have an optional param to create roles for generations other than the latest

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,113 @@
...
+    }
+
+    fun createNewRole(dto: RoleDto): Role {
+        if (roleRepository.findByName(dto.name) != null) {
+            throw IllegalArgumentException(ErrorMessages.roleAlreadyExists)
+        }
```
BrunoRosendo: This isn't right. We can have roles with the same name as long as they're in different generations. I created a role `NoDuplicateRoles` In the generation model but it's only being validated when creating generations. Perhaps we can validate the generation again or port the logic somewhere else. What do you think?

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,113 @@
...
+        generationRepository.save(role.generation)
+        roleRepository.delete(role)
+    }
+
+    fun addUserToRole(roleId: Long, userId: Long) {
+        val role = getRole(roleId)
+        val account = accountService.getAccountById(userId)
+        role.accounts.find { it.id == account.id }.let {
+            if (it != null) throw NoSuchElementException(ErrorMessages.userAlreadyHasRole(roleId, userId))
+        }
```
BrunoRosendo: In my opinion, this operation and the `removeUserFromRole` should be idempotent, thus returning 200 if the user already has the role. Do you find any helpful use case for this error?

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
BrunoRosendo: I'm bumping this comment, it'd be super useful for the frontend devs to have a map of the permissions and respective integers

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
BrunoRosendo: also bump

BrunoRosendo (changes requested):
I ignored the test file and decided to review only the feature code this time, as to not give an over extensive review

BrunoRosendo:
@LuisDuarte1 I found a way to remove all the bloat we have in the services when we would need to define a 2-way relationship for testing purposes. For example, instead of having this:
```kotlin
val activity = activityService.getActivityById(activityId)
val role = getRole(roleId)
val foundPerActivityRole = activity.associatedRoles
        .find { it.activity == activity } ?: PerActivityRole(Permissions())
foundPerActivityRole.role = role
foundPerActivityRole.activity = activity

foundPerActivityRole.permissions.addAll(permissions)
perActivityRoleRepository.save(foundPerActivityRole)

if (activity.associatedRoles.find { it.activity == activity } == null) {
        role.associatedActivities.add(foundPerActivityRole)
}
roleRepository.save(role)
```

We only really need this:
```kotlin
val activity = activityService.getActivityById(activityId)
val role = getRole(roleId)
val foundPerActivityRole = activity.associatedRoles
    .find { it.activity == activity } ?: PerActivityRole(Permissions())
foundPerActivityRole.role = role
foundPerActivityRole.activity = activity

foundPerActivityRole.permissions.addAll(permissions)
perActivityRoleRepository.save(foundPerActivityRole)
```

The second version used to work well in production but failed during tests because the transactions weren't committed until the test method was over. For that reason, I created an utility method to start a new transaction:
```kotlin
fun startNewTransaction() {
    TestTransaction.flagForCommit() // this could be replaced with @Commit annotation on the test method/class
    TestTransaction.end()
    TestTransaction.start()
}
```

Then, whenever we want to check for repository changes after relationships are applied, we only need to call that method to commit the transaction and start a new one:
```kotlin
@Test
        fun `should add permission to role activity and create PerActivityRole`() {
            mockMvc.perform(
                post("/roles/{id}/activities/{activityId}/permissions", testRole.id, project.id)
                    .contentType(MediaType.APPLICATION_JSON)
                    .content(
                        objectMapper.writeValueAsString(
                            mapOf("permissions" to Permissions(listOf(Permission.EDIT_ACTIVITY)))
                        )
                    )
            ).andExpectAll(
                status().isOk
            ).andDocumentEmptyObjectResponse(
                documentationPermissions,
                "Adds an permission to a role activity",
                "It will create a PerRoleActivity if it doesn't exist",
                hasRequestPayload = true,
                urlParameters = parameters
            )

            TestUtils.startNewTransaction()

            assert(roleRepository.findByIdOrNull(testRole.id!!)!!.associatedActivities.size == 1)
            assert(
                roleRepository.findByIdOrNull(testRole.id!!)!!.associatedActivities[0].permissions.contains(
                    Permission.EDIT_ACTIVITY
                )
            )
        }
```

Please also change this in the generation service

BrunoRosendo:
> @LuisDuarte1 I found a way to remove all the bloat we have in the services when we would need to define a 2-way relationship for testing purposes. For example, instead of having this:
> 
> ```kotlin
> val activity = activityService.getActivityById(activityId)
> val role = getRole(roleId)
> val foundPerActivityRole = activity.associatedRoles
>         .find { it.activity == activity } ?: PerActivityRole(Permissions())
> foundPerActivityRole.role = role
> foundPerActivityRole.activity = activity
> 
> foundPerActivityRole.permissions.addAll(permissions)
> perActivityRoleRepository.save(foundPerActivityRole)
> 
> if (activity.associatedRoles.find { it.activity == activity } == null) {
>         role.associatedActivities.add(foundPerActivityRole)
> }
> roleRepository.save(role)
> ```
> 
> We only really need this:
> 
> ```kotlin
> val activity = activityService.getActivityById(activityId)
> val role = getRole(roleId)
> val foundPerActivityRole = activity.associatedRoles
>     .find { it.activity == activity } ?: PerActivityRole(Permissions())
> foundPerActivityRole.role = role
> foundPerActivityRole.activity = activity
> 
> foundPerActivityRole.permissions.addAll(permissions)
> perActivityRoleRepository.save(foundPerActivityRole)
> ```
> 
> The second version used to work well in production but failed during tests because the transactions weren't committed until the test method was over. For that reason, I created an utility method to start a new transaction:
> 
> ```kotlin
> fun startNewTransaction() {
>     TestTransaction.flagForCommit() // this could be replaced with @Commit annotation on the test method/class
>     TestTransaction.end()
>     TestTransaction.start()
> }
> ```
> 
> Then, whenever we want to check for repository changes after relationships are applied, we only need to call that method to commit the transaction and start a new one:
> 
> ```kotlin
> @Test
>         fun `should add permission to role activity and create PerActivityRole`() {
>             mockMvc.perform(
>                 post("/roles/{id}/activities/{activityId}/permissions", testRole.id, project.id)
>                     .contentType(MediaType.APPLICATION_JSON)
>                     .content(
>                         objectMapper.writeValueAsString(
>                             mapOf("permissions" to Permissions(listOf(Permission.EDIT_ACTIVITY)))
>                         )
>                     )
>             ).andExpectAll(
>                 status().isOk
>             ).andDocumentEmptyObjectResponse(
>                 documentationPermissions,
>                 "Adds an permission to a role activity",
>                 "It will create a PerRoleActivity if it doesn't exist",
>                 hasRequestPayload = true,
>                 urlParameters = parameters
>             )
> 
>             TestUtils.startNewTransaction()
> 
>             assert(roleRepository.findByIdOrNull(testRole.id!!)!!.associatedActivities.size == 1)
>             assert(
>                 roleRepository.findByIdOrNull(testRole.id!!)!!.associatedActivities[0].permissions.contains(
>                     Permission.EDIT_ACTIVITY
>                 )
>             )
>         }
> ```
> 
> Please also change this in the generation service

Note that this only occurs because we're using `@Transactional` in the test method. This test specifically works fine without any of that. This thread shows an alternative to using `@Transactional` in the accepted answer: https://stackoverflow.com/questions/44079793/how-to-force-transaction-commit-in-spring-boot-test

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: I thought of it being useful for displaying an error in the frontend if the user is already in role (maybe I should change the exception to make it a more useful HTTP code), it can also be idempotent but this verification must be then be done on the client side which adds a bit of unnecessary complexity, wdyt?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
LuisDuarte1: Nice catch, will do!

src/main/kotlin/pt/up/fe/ni/website/backend/dto/permissions/UserIdDto.kt
LuisDuarte1: Done

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
LuisDuarte1: Just to make sure, I should edit the RoleDTO to add a nullable generation id, right?

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: We can't use `@Valid` on expressions so I'm not sure how I would reuse `NoDuplicateRoles` in this case, maybe we should port the logic somewhere else to make it more readable, not sure where i'd put it since I don't like utility classes very much :smile:

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
BrunoRosendo: Yes that should work just fine. The fields that don't correspond to the entity should be ignored when creating roles so it should work

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
BrunoRosendo: @LuisDuarte1 
- 1st option: Check `EntityDto`and see how we use the validator. We can use the same logic to validate the generation once the role is added. You can either validate before saving or let Spring rollback the changes when the exception is thrown (I think this is the behaviour since we're using @Transactional but please check)
- 2nd option: I don't see a good way tbh I don't like utils classes either but it's hard to find a better option than the 1st above. We could theoretically access the `isValid` method of `NoDuplicateRolesValidator` but that's cursed. The alternative would be a class with validation methods or repeating the 2 lines of code.

Imo the 1st option is cleaner

BrunoRosendo: I didn't test it but it seems possible to validate a single field if you want: https://codingexplained.com/coding/java/hibernate/validate-single-field-hibernate-validator . However, validating the entire generation shouldn't bring any problems

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
BrunoRosendo: I see 2 cases in the frontend:
- The user already has a role so the UI would be displaying the right info and the request wouldn't happen
- The UI is out of sync (e.g. bad network, someone else updated it) so the FE requests to add a role, receives 200 and updates the UI accordingly. Nothing really changes in the server but the result is as expected for the API user.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
limwa: @LuisDuarte1 this conversation has been marked as resolved but from looking at the RoleDto.kt file, it still hasn't been changed, has this been done on purpose?

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
limwa: I think so too!

src/main/kotlin/pt/up/fe/ni/website/backend/repository/RoleRepository.kt
```diff
@@ -5,4 +5,6 @@ import org.springframework.stereotype.Repository
 import pt.up.fe.ni.website.backend.model.Role
 
 @Repository
-interface RoleRepository : CrudRepository<Role, Long>
+interface RoleRepository : CrudRepository<Role, Long> {
+    fun findByName(name: String): Role?
```
limwa: Since name isn't unique, is there a way to specify that we only want the role associated with the most recent generation?

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,113 @@
...
-        foundActivity.activity = activity
+        foundPerActivityRole.role = role
+        foundPerActivityRole.activity = activity
 
-        foundActivity.role = role
-        foundActivity.permissions.add(permission)
-        perActivityRoleRepository.save(foundActivity)
+        foundPerActivityRole.permissions.addAll(permissions)
+        perActivityRoleRepository.save(foundPerActivityRole)
+        if (activity.associatedRoles.find { it.activity == activity } == null) {
```
limwa: This check is already performed above. I'd suggest extracting it to a variable and use it on line 47 and 53!

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,113 @@
...
+        role.accounts.add(account)
+        roleRepository.save(role)
+    }
+
+    fun removeUserFromRole(roleId: Long, userId: Long) {
+        val role = getRole(roleId)
+        val account = accountService.getAccountById(userId)
+        role.accounts.find { it.id == account.id }.let {
+            if (it == null) throw NoSuchElementException(ErrorMessages.userNotInRole(roleId, userId))
+        }
```
limwa:
```kotlin
if (role.accounts.none { it.id == account.id }) {
    throw NoSuchElementException(ErrorMessages.userNotInRole(roleId, userId))
}
```
I think this makes the code a bit more readable!

limwa (changes requested):
Overall, the PR looks pretty good, so well done! Most of my suggestions are related to code readability so they should be quick to go through.

src/main/kotlin/pt/up/fe/ni/website/backend/repository/RoleRepository.kt
LuisDuarte1: I don't think it is possible without a custom query, but it only is used on role creation I think, but I'll change to make the according changes that @BrunoRosendo suggested and therefore I don't think findByName will be necessary at all.

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: Done ^^

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
LuisDuarte1: Sorry I forgot to commit the changes :sweat_smile:

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: I agree, I'll change them to make them idempotent

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: In order to make the this api call idempotent according to @BrunoRosendo suggestion ~I'll have to change this to:
`if(role.accounts.none { it.id == account.id }) return`~

Edit: This check is not even needed for deleting the role since `.remove` does nothing if the account doesnt exist ^^

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: I can now remove this logic because it was due to ManyToOne transaction shenanigans on the test enviroment that can be fixed with the new function `TestUtils.startNewTransaction()`

limwa: Nice!

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
LuisDuarte1: I think I'll need @Sirze01 for this one :sweat_smile: ahahha not sure how I would do it

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
LuisDuarte1: I've changed it, it's more clear, but probably not clear enough ahahah idk how to describe it well enough

LuisDuarte1:
Just a quick status update:

- I've removed the OrderColumn on the Activity model because it was causing the same error as #172 and should be also solved in a later PR
- The `TestUtils.startNewTransaction()` idea works almost flawlessly except you can't mark the transaction for commit if the transaction fails (this is not ideal in testing because we should test all cases) so, I've created the `rollback`flag that marks the transaction if it's expected to fail.
- I've not updated all tests to use `TestUtils.startNewTransaction()` just the ones that need the 2-way relationship for assertions (I think updating generations test cases should be done seperately)
- You can now specify a generationId on role creation for creating roles for the past years, if not specified it will default to the latest one.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/RoleController.kt
limwa: @LuisDuarte1 this still needs to be changed

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
limwa: This also needs to be changed

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
limwa: @LuisDuarte1 this also needs to be done

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,110 @@
...
+    fun createNewRole(dto: RoleDto): Role {
+        val role = dto.create()
+        val generation: Generation = if (dto.generationId != null) {
+            generationRepository.findById(dto.generationId).orElseThrow {
+                IllegalArgumentException(ErrorMessages.generationNotFound(dto.generationId))
+            }
+        } else {
+            generationRepository.findFirstByOrderBySchoolYearDesc()
+                ?: throw IllegalArgumentException(ErrorMessages.noGenerations)
+        }
```
limwa: We'll end up using this a lot, so I think it should be extracted to it's own function!

limwa (changes requested):
Almost there! I just marked some comments that haven't been properly dealt with and one other suggestions!

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadRoles.kt
BrunoRosendo: You may want to open an issue for that and CC him

BrunoRosendo:
> 

Please change this in the generation service too: https://github.com/NIAEFEUP/website-niaefeup-backend/blob/0c670a9d8486d1441fc1da7adefc626a0bf015df/src/main/kotlin/pt/up/fe/ni/website/backend/service/GenerationService.kt#L92

I commented the lines where I defined both sides of the relationship

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,110 @@
...
-        foundActivity.permissions.add(permission)
-        perActivityRoleRepository.save(foundActivity)
+        perActivityRole.permissions.addAll(permissions)
+        perActivityRoleRepository.save(perActivityRole)
     }
 
-    fun revokePermissionFromRoleOnActivity(role: Role, activity: Activity, permission: Permission) {
+    fun revokePermissionFromRoleOnActivity(roleId: Long, activityId: Long, permissions: Permissions) {
+        val activity = activityService.getActivityById(activityId)
+        val role = getRole(roleId)
```
BrunoRosendo: We could just use the ID for comparison but the performance difference should be negligible

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
BrunoRosendo: I agree!

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,110 @@
...
+            generationRepository.findById(dto.generationId).orElseThrow {
+                IllegalArgumentException(ErrorMessages.generationNotFound(dto.generationId))
+            }
+        } else {
+            generationRepository.findFirstByOrderBySchoolYearDesc()
+                ?: throw IllegalArgumentException(ErrorMessages.noGenerations)
+        }
+
+        role.generation = generation
+        generation.roles.add(role)
```
BrunoRosendo: Only the generation side is necessary

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,110 @@
...
+        val validator: Validator = ApplicationContextUtils.getBean(Validator::class.java)
```
BrunoRosendo: Don't request beans every time you run the function. Inject it in the controller, the same way we do for repositories (they're also beans).

We also don't need to use `ApplicationContextUtils`, we only use that in EntityDto because DTOs are not Spring components

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,110 @@
...
-        foundActivity.permissions.remove(permission)
+        foundActivity.permissions.removeAll(permissions)
         perActivityRoleRepository.save(foundActivity)
     }
+
+    fun createNewRole(dto: RoleDto): Role {
+        val role = dto.create()
+        val generation: Generation = if (dto.generationId != null) {
+            generationRepository.findById(dto.generationId).orElseThrow {
+                IllegalArgumentException(ErrorMessages.generationNotFound(dto.generationId))
```
BrunoRosendo:
```suggestion
                NoSuchElementException(ErrorMessages.generationNotFound(dto.generationId))
```
404 makes more sense, do you agree?

BrunoRosendo (changes requested):
I just noticed we don't have a way of updating some role fields, such as name and isSection. We can discuss that in the next meeting

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -1,46 +1,110 @@
...
+                ?: throw IllegalArgumentException(ErrorMessages.noGenerations)
+        }
+
+        role.generation = generation
+        generation.roles.add(role)
+        val validator: Validator = ApplicationContextUtils.getBean(Validator::class.java)
+        // we can infer that if something goes wrong is with adding a new role with the same name
+        if (validator.validate(generation).isNotEmpty()) {
+            throw IllegalArgumentException(ErrorMessages.roleAlreadyExists(role.name, generation.schoolYear))
+        }
```
BrunoRosendo: This is true if nothing else has gone wrong in other requests but sure.

But I do prefer validating just the roles property, `validator.validateProperty(generation, "roles")`

src/test/kotlin/pt/up/fe/ni/website/backend/utils/TestUtils.kt
```diff
@@ -11,5 +12,15 @@ class TestUtils {
                 .apply { set(year, month, day, 0, 0, 0) }
                 .time
         }
+
+        fun startNewTransaction(rollback: Boolean = false) {
+            if (rollback) {
```
BrunoRosendo, line 17: nice touch

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: This is done because `getRoleById` checks if the role exist and throws if it doesnt

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: This is already done, the optional generation id is inside the `roleDto`

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
BrunoRosendo: Makes sense

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: done

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: done

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: nice catch, done

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: without the generation side it doesn't work for some reason :/

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: done

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: done

limwa (approved):
LGTM 🚀

BrunoRosendo:
> I just noticed we don't have a way of updating some role fields, such as name and isSection. We can discuss that in the next meeting

@LuisDuarte1 I only want to discuss this in a meeting, I forgot last time

LuisDuarte1:
> I just noticed we don't have a way of updating some role fields, such as name and isSection. 

Done

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
```diff
@@ -98,4 +99,17 @@ class RoleService(
         role.accounts.remove(account)
         roleRepository.save(role)
     }
+
+    fun updateRole(roleId: Long, dto: UpdateRoleDto): Role {
+        val role = getRoleById(roleId)
+
+        // just for validation
+        role.name = dto.name
```
BrunoRosendo: why don't you call `dto.update(role)` and then validate before saving?

BrunoRosendo (approved):

src/main/kotlin/pt/up/fe/ni/website/backend/service/RoleService.kt
LuisDuarte1: thats an oversight of mine, done!

PR #148, opened 2023-04-20
Fixed startup error related to final Activity fields

BrunoRosendo:
There was an error when booting Spring because abstract models apparently cannot contain final getters, This PR changes this behaviour

# Review checklist


BrunoRosendo:
@bdmendes Given the error, should we reconsider this plugin? https://kotlinlang.org/docs/all-open-plugin.html#command-line-compiler

MRita443 (approved):

bdmendes:
> @bdmendes Given the error, should we reconsider this plugin? https://kotlinlang.org/docs/all-open-plugin.html#command-line-compiler

I think it is still a very specific use case, so let's skip it for now.

bdmendes (approved):

PR #152, opened 2023-04-20
CORS config

### LINKED ISSUE #58: security: use proper cors in security config

jamcunha:
Closes #58 

# Review checklist


src/main/resources/application.properties
```diff
@@ -35,6 +35,9 @@ upload.static-path=classpath:static
 # URL that will serve static content
 upload.static-serve=http://localhost:3000/static
 
+# Frontend Url
+frontend.url = http://localhost:3000
```
BrunoRosendo: Really small detail but origin probably makes more sense than url. Perhaps it's even better to call it something like cors.allow-origin since it's the only use for the origin

src/main/kotlin/pt/up/fe/ni/website/backend/config/CorsConfig.kt
```diff
@@ -0,0 +1,31 @@
+package pt.up.fe.ni.website.backend.config
```
BrunoRosendo: Just feels a bit weird having this at the root of config, since we have packages for other configs. I'll suggest moving this into `config/auth` and renaming the package to something like `security`

BrunoRosendo (changes requested):

BrunoRosendo (approved):
good job

bdmendes (approved):

PR #180, opened 2023-07-22
Refactor/rest like endpoints

### LINKED ISSUE #143: Refactor endpoints to follow a more "REST-like" approach

Sirze01:
Closes #143 

First of all kudos to @BrunoRosendo for the great groundwork.
This PR refactors some of the endpoints being used and adds support for URL queries in the documentation, as, as of now, some endpoints rely on URL queries to filter the returned contents:

Next is a list of the affected endpoints:
- accounts
    - POST /changePassord/{id} -> POST /{id}/password
    - POST /new -> POST

- auth
    - POST /new -> POST

- events
    - GET /category -> GET ?category=Great Events (Missing documentation for query parameters)
    - POST /new -> POST
    - PUT /{idEvent}/addTeamMember/{idAccount} -> PUT /{idEvent}/team/{idAccount}
    - PUT /{idEvent}/removeTeamMember/{idAccount} -> DELETE /{idEvent}/team/{idAccount}

- generations
    - GET -> GET /years
    - POST /new -> POST

- posts
    - POST new -> POST

- projects
    - POST /new -> POST
    - PUT /{idProject}/addTeamMember/{idAccount} -> PUT /{idProject}/team/{idAccount}
    - PUT /{idProject}/removeTeamMember/{idAccount} -> DELETE /{idProject}/team/{idAccount}
    - PUT /{idEvent}/addHallOffFameMember/{idAccount} -> PUT /{idEvent}/hallOfFameMember/{idAccount}
    - PUT /{idEvent}/removeHallOfFameMember/{idAccount} -> DELETE /{idEvent}/hallOfFameMember/{idAccount}

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -30,7 +30,7 @@ class ProjectController(private val service: ProjectService) {
     @GetMapping("/{projectSlug}**")
     fun getProjectBySlug(@PathVariable projectSlug: String) = service.getProjectBySlug(projectSlug)
 
-    @PostMapping("/new", consumes = ["multipart/form-data"])
+    @PostMapping(consumes = ["multipart/form-data"])
     fun createProject(
         @RequestPart project: ProjectDto,
         @RequestParam
@@ -65,13 +65,13 @@ class ProjectController(private val service: ProjectService) {
     @PutMapping("/{id}/unarchive")
     fun unarchiveProjectById(@PathVariable id: Long) = service.unarchiveProjectById(id)
 
-    @PutMapping("/{idProject}/addTeamMember/{idAccount}")
+    @PutMapping("/{idProject}/team/{idAccount}")
     fun addTeamMemberById(
         @PathVariable idProject: Long,
         @PathVariable idAccount: Long
     ) = service.addTeamMemberById(idProject, idAccount)
 
-    @PutMapping("/{idProject}/removeTeamMember/{idAccount}")
+    @DeleteMapping("/{idProject}/team/{idAccount}")
     fun removeTeamMemberById(
         @PathVariable idProject: Long,
         @PathVariable idAccount: Long
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -77,13 +77,13 @@ class ProjectController(private val service: ProjectService) {
         @PathVariable idAccount: Long
     ) = service.removeTeamMemberById(idProject, idAccount)
 
-    @PutMapping("/{idProject}/hallOfFameMember/{idAccount}")
+    @PutMapping("/{idProject}/hallOfFame/{idAccount}")
     fun addHallOfFameMemberById(
         @PathVariable idProject: Long,
         @PathVariable idAccount: Long
     ) = service.addHallOfFameMemberById(idProject, idAccount)
 
-    @DeleteMapping("/{idProject}/hallOfFameMember/{idAccount}")
+    @DeleteMapping("/{idProject}/hallOfFame/{idAccount}")
     fun removeHallOfFameMemberById(
         @PathVariable idProject: Long,
         @PathVariable idAccount: Long
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
@@ -77,13 +77,13 @@ class ProjectController(private val service: ProjectService) {
         @PathVariable idAccount: Long
     ) = service.removeTeamMemberById(idProject, idAccount)
 
-    @PutMapping("/{idProject}/addHallOfFameMember/{idAccount}")
+    @PutMapping("/{idProject}/hallOfFameMember/{idAccount}")
     fun addHallOfFameMemberById(
         @PathVariable idProject: Long,
         @PathVariable idAccount: Long
     ) = service.addHallOfFameMemberById(idProject, idAccount)
 
-    @PutMapping("/{idProject}/removeHallOfFameMember/{idAccount}")
+    @DeleteMapping("/{idProject}/hallOfFameMember/{idAccount}")
     fun removeHallOfFameMemberById(
         @PathVariable idProject: Long,
         @PathVariable idAccount: Long
--- a/src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
+++ b/src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
@@ -10,6 +10,7 @@ import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.mockito.Mockito
 import org.springframework.beans.factory.annotation.Autowired
@@ -27,7 +28,6 @@ import pt.up.fe.ni.website.backend.model.CustomWebsite
 import pt.up.fe.ni.website.backend.model.Event
 import pt.up.fe.ni.website.backend.model.constants.ActivityConstants
 import pt.up.fe.ni.website.backend.model.constants.EventConstants as Constants
-import org.junit.jupiter.api.Nested
 import pt.up.fe.ni.website.backend.model.embeddable.DateInterval
 import pt.up.fe.ni.website.backend.repository.AccountRepository
 import pt.up.fe.ni.website.backend.repository.EventRepository
@@ -149,7 +149,6 @@ internal class EventControllerTest @Autowired constructor(
                 )
             )
             extraEvents.forEach { repository.save(it) }
-
 
             mockMvc.perform(get("/events?category={category}", testEvent.category))
                 .andExpectAll(
--- a/src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
+++ b/src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
@@ -27,6 +27,7 @@ import pt.up.fe.ni.website.backend.model.CustomWebsite
 import pt.up.fe.ni.website.backend.model.Event
 import pt.up.fe.ni.website.backend.model.constants.ActivityConstants
 import pt.up.fe.ni.website.backend.model.constants.EventConstants as Constants
+import org.junit.jupiter.api.Nested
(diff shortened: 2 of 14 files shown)


Sirze01:
Some endpoints are somewhat rule-breakers.
I thought of some alternatives, if you guys think we should change them, let me know. 

I think having different operations to handle the team and project memberships is somewhat of an anti-pattern in REST terms. Having the operations seems to be way too imperative, almost resembling Remote Procedure Calls (RPC).

- Events
  - Single `PUT /{idEvent}/team` (with a list "team" resource for simplification), the list of members would be rewritten with every update
  - Patching the Event object with the new team members list
  - Just rewriting the whole event object

- Projects
  - Patch to the project object with the new value of `isArchived`
  - `PUT /{id}/` with the modified body in the case of the  archival/unarchival

This is me being pedantic at this point, as I see value in these exceptions, making the endpoints easier to interact with.
Let me know your opinion.

BrunoRosendo:
I haven't reviewed but I want to warn that after #139 this PR needs to address the API on the new endpoints. Let's discuss the team decision in the next meeting

LuisDuarte1:
> Some endpoints are somewhat rule-breakers. I thought of some alternatives, if you guys think we should change them, let me know.
> 
> I think having different operations to handle the team and project memberships is somewhat of an anti-pattern in REST terms. Having the operations seems to be way too imperative, almost resembling Remote Procedure Calls (RPC).
> 
> * Events
>   
>   * Single `PUT /{idEvent}/team` (with a list "team" resource for simplification), the list of members would be rewritten with every update
>   * Patching the Event object with the new team members list
>   * Just rewriting the whole event object
> * Projects
>   
>   * Patch to the project object with the new value of `isArchived`
>   * `PUT /{id}/` with the modified body in the case of the  archival/unarchival
> 
> This is me being pedantic at this point, as I see value in these exceptions, making the endpoints easier to interact with. Let me know your opinion.

I think that it's not pedantic at all and you approach is simpler imo, it simplifies the backend by reducing the number of endpoints, by having the side-effect of sometimes making more api calls (eg.: to add a team member, you have to get the current list and then make the request to rewrite the list of members with the new member. In most situations there's no additional overhead, but in some cases it'll add a request of overhead). I think we need more opinion on this cc: @MRita443

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -112,14 +112,94 @@ internal class EventControllerTest @Autowired constructor(
...
+                        |It also allows to filter the events by category, using the query parameter "category".
                         |This is useful for example in the frontend's event page, where events are displayed.
                     """.trimMargin()
                 )
         }
     }
 
+    @NestedTest
+    @DisplayName("GET events?category={category}")
+    inner class GetEventsByCategory {
```
BrunoRosendo: I think all these tests should be moved into `GetEvents`

BrunoRosendo (changes requested):
Good job with the refactor!

Thoughts/Answers to your comment:
- As for adding members to the teams, I think we could change to POST /events/{id}/team and pass an id in the body, resembling the creation of a new team member. I think this is more RESTful and keeps the API simpler to use
- As for project archival/unarchival, if we want to be very strict about REST norms then we could erase it since we can change it using the regular PUT. The reason I proposed these endpoints at the time was that I could see a clear purpose for them (e.g. clicking an archive button instead of editing a form that updates the project). We can also say that updating and archiving entities are different things (independently of implementation), there seems to be a discussion on this https://stackoverflow.com/questions/16201905/restful-archiving-of-entities-in-webapi
Personaly, I prefer having an exception to REST and keeping the API easily understandable, just like [GitHub](https://docs.github.com/en/free-pro-team@latest/rest/issues/issues?apiVersion=2022-11-28#lock-an-issue) and [Reddit](https://www.reddit.com/dev/api/#POST_api_mod_conversations_:conversation_id_archive) do in some cases.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
BrunoRosendo: This got me thinking that we might want to have more filters on the BE (e.g. archived posts) but I think we can wait to see how FE wants to implement filters

Sirze01: > I think all these tests should be moved into `GetEvents`

Will do!

Sirze01:
> Good job with the refactor!
> 
> Thoughts/Answers to your comment:
> 
>     * As for adding members to the teams, I think we could change to POST /events/{id}/team and pass an id in the body, resembling the creation of a new team member. I think this is more RESTful and keeps the API simpler to use
> 
>     * As for project archival/unarchival, if we want to be very strict about REST norms then we could erase it since we can change it using the regular PUT. The reason I proposed these endpoints at the time was that I could see a clear purpose for them (e.g. clicking an archive button instead of editing a form that updates the project). We can also say that updating and archiving entities are different things (independently of implementation), there seems to be a discussion on this https://stackoverflow.com/questions/16201905/restful-archiving-of-entities-in-webapi
>       Personaly, I prefer having an exception to REST and keeping the API easily understandable, just like [GitHub](https://docs.github.com/en/free-pro-team@latest/rest/issues/issues?apiVersion=2022-11-28#lock-an-issue) and [Reddit](https://www.reddit.com/dev/api/#POST_api_mod_conversations_:conversation_id_archive) do in some cases.

- Regarding team updates
I would go even further and instead of taking only a member ID in the body, take an array of IDs to support batch updates, (Adding/removing more than one in a "commit" action in the frontend)

- Regarding the archive/unarchive
I think we should leave the endpoints as they are now.

BrunoRosendo:
#64 was merged, could you please refactor the endpoints too? It's the exact same thing as adding team members

LuisDuarte1:
@Sirze01 can you look into this please?

Sirze01:
> Closes #143
> 
> First of all kudos to @BrunoRosendo for the great groundwork. This PR refactors some of the endpoints being used and adds support for URL queries in the documentation, as, as of now, some endpoints rely on URL queries to filter the returned contents:
> 
> Next is a list of the affected endpoints:
> 
>     * accounts
>       
>       * POST /changePassord/{id} -> POST /{id}/password
>       * POST /new -> POST
> 
>     * auth
>       
>       * POST /new -> POST
> 
>     * events
>       
>       * GET /category -> GET ?category=Great Events (Missing documentation for query parameters)
>       * POST /new -> POST
>       * PUT /{idEvent}/addTeamMember/{idAccount} -> PUT /{idEvent}/team/{idAccount}
>       * PUT /{idEvent}/removeTeamMember/{idAccount} -> DELETE /{idEvent}/team/{idAccount}
> 
>     * generations
>       
>       * GET -> GET /years
>       * POST /new -> POST
> 
>     * posts
>       
>       * POST new -> POST
> 
>     * projects
>       
>       * POST /new -> POST
>       * PUT /{idProject}/addTeamMember/{idAccount} -> PUT /{idProject}/team/{idAccount}
>       * PUT /{idProject}/removeTeamMember/{idAccount} -> DELETE /{idProject}/team/{idAccount}
>       * PUT /{idEvent}/addHallOffFameMember/{idAccount} -> PUT /{idEvent}/hallOfFameMember/{idAccount}
>       * PUT /{idEvent}/removeHallOfFameMember/{idAccount} -> DELETE /{idEvent}/hallOfFameMember/{idAccount}
> 
> 
> # Review checklist
> 
>     * [x]  Properly documents API changes in the corresponding controller test
> 
>     * [x]  Contains enough appropriate tests
> 
>     * [x]  Behavior is as expected
> 
>     * [x]  Clean, well structured code

Updated the PR description.
The [PR in the frontend seems to be updated still](https://github.com/NIAEFEUP/website-niaefeup-frontend/pull/219).

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -65,25 +65,25 @@ class ProjectController(private val service: ProjectService) {
...
 
-    @PutMapping("/{idProject}/removeTeamMember/{idAccount}")
+    @DeleteMapping("/{idProject}/team/{idAccount}")
     fun removeTeamMemberById(
         @PathVariable idProject: Long,
         @PathVariable idAccount: Long
     ) = service.removeTeamMemberById(idProject, idAccount)
 
-    @PutMapping("/{idProject}/addHallOfFameMember/{idAccount}")
+    @PutMapping("/{idProject}/hallOfFameMember/{idAccount}")
```
BrunoRosendo:
```suggestion
    @PutMapping("/{idProject}/hallOfFame/{idAccount}")
```
I think this is more similar to "team", wdyt?

BrunoRosendo (changes requested):
really small detail!

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
Sirze01: Good catch. Already updated it, take a look.

BrunoRosendo (approved):
Great job!

LuisDuarte1 (approved):
LGTM, great work :rocket:

PR #181, opened 2023-07-25
Fixed unchanged email account updates

### LINKED ISSUE #176: accounts: wrong error when updating
This comes from the fact that our updates are creations (as PUTs) of all the information of the entity. This means that, in order to update an account, we need send all of its fields including the email.

Currently, we're checking for email duplicates without taking this into consideration, so it's impossible to update the account with the same email:
```kotlin
// In AccountService.kt
fun updateAccountById(id: Long, dto: UpdateAccountDto): Account {
        val account = getAccountById(id)

        repository.findByEmail(dto.email)?.let {
            throw IllegalArgumentException(ErrorMessages.emailAlreadyExists)
        }

        dto.photoFile?.let {
            val fileName = photoFilename(dto.email, it)
        }

        val newAccount = dto.update(account)

        return repository.save(newAccount)
    }
 ```

We need to change this to not throw an error if the email is the same as before

MRita443:
Closes #176 

Checks if the email was changed before comparing with already used emails. Adds a test case to verify this instance.

# Review checklist

### DIFF
--- a/src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
+++ b/src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
@@ -48,8 +48,10 @@ class AccountService(
     fun updateAccountById(id: Long, dto: UpdateAccountDto): Account {
         val account = getAccountById(id)
 
-        repository.findByEmail(dto.email)?.let {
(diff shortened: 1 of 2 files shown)


BrunoRosendo (approved):
Perfect!

pedrosilva17 (approved):
No complaints. Good work 👍

Issue #144, opened 2023-03-29
docs: documentation for multi part form


bdmendes:
After #71, we need to update our documentation to reflect the new changes. This wasn't done in the PR because we need to add support for multi part form docs.

BrunoRosendo:
We will also need this for projects and events, after #178

Isaac3105:
As we are shifting towards building a new backend from zero this issue will be abandoned.

PR #203, opened 2023-11-21
43 events gallery of photos

### LINKED ISSUE #43: events: gallery of photos
An event (and maybe a project too) should have a gallery displaying an arbitrary number of photos of the event after it occurred.

We should add the path of these photos to the model and create endpoints capable of storing and deleting those photos. We should also ensure the `GET project/{id}` endpoint includes this gallery.

AvilaAndre:
Closes #43 

Creates the endpoints to manage events' photos.

# Review checklist


MRita443:
Would it be possible to generalize the gallery to activity? We were discussing this today and think that projects benefit from a gallery too for all the images in their details page

AvilaAndre:
> Would it be possible to generalize the gallery to activity? We were discussing this today and think that projects benefit from a gallery too for all the images in their details page

Of course! I'll get onto it :)

MRita443:
Hi @AvilaAndre are the alterations now ready?

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
```diff
@@ -43,6 +43,8 @@ abstract class Activity(
     @field:NotBlank
     open var image: String,
 
+    val gallery: MutableList<String> = mutableListOf(),
```
MRita443:
```suggestion
    @ElementCollection
    open val gallery: MutableList<String> = mutableListOf(),
```

MRita443: To tackle the `Basic attribute type should not be a container.` error

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -83,4 +84,35 @@ abstract class AbstractActivityService<T : Activity>(
         activity.teamMembers.removeIf { it.id == idAccount }
         return repository.save(activity)
     }
+
+    fun addGalleryPhoto(activityId: Long, image: MultipartFile): Activity {
+        if (!repository.existsById(activityId)) {
+            throw NoSuchElementException(ErrorMessages.eventNotFound(activityId))
```
MRita443: Here you are throwing a message about an `Event` instead of an `Activity`. If your intention is to verify if the `Activity` exists, the `getActivityById(activityId)` already does that

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -83,4 +84,35 @@ abstract class AbstractActivityService<T : Activity>(
...
+
+        activity.gallery.add(imageName)
+
+        return repository.save(activity)
+    }
+
+    fun removeGalleryPhoto(activityId: Long, photoName: String): Activity {
+        if (!repository.existsById(activityId)) {
+            throw NoSuchElementException(ErrorMessages.eventNotFound(activityId))
+        }
```
MRita443: Same as above

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
```diff
@@ -44,6 +44,8 @@ object ErrorMessages {
 
     fun roleNotFound(id: Long): String = "role not found with id $id"
 
+    fun photoNotFound(): String = "photo not found"
```
MRita443: I think it would be beneficial to include the image's name here, and perhaps make it more general by using "Image" instead of "Photo"

src/main/kotlin/pt/up/fe/ni/website/backend/controller/EventController.kt
```diff
@@ -74,4 +74,18 @@ class EventController(private val service: EventService) {
         @PathVariable idEvent: Long,
         @PathVariable idAccount: Long
     ) = service.removeTeamMemberById(idEvent, idAccount)
+
+    @PutMapping("/{idEvent}/gallery/addPhoto", consumes = ["multipart/form-data"])
```
MRita443: @Sirze01 Could you verify the restfulness of these endpoints ;P?

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadActivity.kt
```diff
@@ -89,6 +89,13 @@ class PayloadActivity {
                     "Description of the event",
                     JsonFieldType.STRING,
                     optional = true
+                ),
+                DocumentedJSONField(
+                    "gallery[]",
+                    "Array of photos associated with the activity",
+                    JsonFieldType.ARRAY,
+                    isInRequest = false,
```
MRita443, line 102: Given how the image of each `Activity` is being documented thus far, I'd leave leave this line out and specify that what is being saved is the path to these images

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -83,4 +84,35 @@ abstract class AbstractActivityService<T : Activity>(
...
+
+    fun addGalleryPhoto(activityId: Long, image: MultipartFile): Activity {
+        if (!repository.existsById(activityId)) {
+            throw NoSuchElementException(ErrorMessages.eventNotFound(activityId))
+        }
+
+        val activity = getActivityById(activityId)
+
+        val fileName = fileUploader.buildFileName(image, activity.title)
+        val imageName = fileUploader.uploadImage("gallery", fileName, image.bytes)
```
MRita443: Here it would be best to prefix this path with the `IMAGE_FOLDER` of each `Activity`, in order to separate them , no? Either that or just that same exact folder, what do you think?

MRita443 (changes requested):
Few suggestions so far, will look at the tests after these changes :)

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
AvilaAndre: In order to use @ElementCollection the fetch type must be changed to Eager because by default it is Lazy, being Lazy the value must be obtained through a query or else it throws an exception, that's why my tests were failing en masse :p

src/main/kotlin/pt/up/fe/ni/website/backend/controller/EventController.kt
```diff
@@ -74,4 +74,18 @@ class EventController(private val service: EventService) {
...
+        @RequestParam
+        @ValidImage
+        image: MultipartFile
+    ) = service.addGalleryImage(idEvent, image)
+
+    @PutMapping("/{idEvent}/gallery/removeImage")
+    fun removeGalleryImage(
+        @PathVariable idEvent: Long,
+        @RequestPart imageUrl: String
+    ) = service.removeGalleryImage(idEvent, imageUrl)
```
BrunoRosendo: As rita pointed out, this should be changed according to #180

src/main/kotlin/pt/up/fe/ni/website/backend/service/ErrorMessages.kt
```diff
@@ -44,6 +44,8 @@ object ErrorMessages {
 
     fun roleNotFound(id: Long): String = "role not found with id $id"
 
+    fun imageNotFound(imageName: String): String = "image not found with name $imageName"
```
BrunoRosendo: I would personally generalize this to `fileNotFound`, so we can use it for other files

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -74,13 +79,40 @@ abstract class AbstractActivityService<T : Activity>(
...
+    fun addGalleryImage(activityId: Long, image: MultipartFile): Activity {
+        val activity = getActivityById(activityId)
+
+        var imageFolder = "activities"
+
+        if (activity is Event) {
+            imageFolder = EventService.IMAGE_FOLDER
+        } else if (activity is Project) {
+            imageFolder = ProjectService.IMAGE_FOLDER
+        }
```
BrunoRosendo: I think this is a big violation of OOP and we can improve this. My suggestion is to a method like this in the abstract activity service:

```suggestion
fun addGalleryImage(activityId: Long, image: MultipartFile, imageFolder: String): Activity {
        val activity = getActivityById(activityId)
        
        val fileName = fileUploader.buildFileName(image, activity.title)
        val imageName = fileUploader.uploadImage(imageFolder + "/gallery", fileName, image.bytes)

        activity.gallery.add(imageName)

        return repository.save(activity)
    }
```

And then add new methods to the project/events services where the name is chosen. This would maintain the logic and decouple the inheritance logic from the abstract service

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -74,13 +79,40 @@ abstract class AbstractActivityService<T : Activity>(
...
+
+        val fileName = fileUploader.buildFileName(image, activity.title)
+        val imageName = fileUploader.uploadImage(imageFolder + "/gallery", fileName, image.bytes)
+
+        activity.gallery.add(imageName)
+
+        return repository.save(activity)
+    }
+
+    fun removeGalleryImage(activityId: Long, imageName: String): Activity {
```
BrunoRosendo, line 97: We should also erase the photo from storage!

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -271,6 +274,89 @@ internal class EventControllerTest @Autowired constructor(
         }
     }
 
+    @NestedTest
+    @DisplayName("GET events/category/{category}")
```
BrunoRosendo, line 278: Was this not tested before? I'm a bit confused since it seems out of scope for this PR

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -725,6 +812,169 @@ internal class EventControllerTest @Autowired constructor(
...
+                .addFile("image", contentType = MediaType.IMAGE_JPEG_VALUE)
+                .perform()
+                .andExpectAll(
+                    status().isOk,
+                    content().contentType(MediaType.APPLICATION_JSON),
+                    jsonPath("$.title").value(testEvent.title),
+                    jsonPath("$.description").value(testEvent.description),
+                    jsonPath("$.teamMembers.length()").value(testEvent.teamMembers.size),
+                    jsonPath("$.gallery.length()").value(1),
+                    jsonPath("$.gallery[0]").value(expectedImagePath),
```
BrunoRosendo, line 857: Have you (manually) tested if you can access this path on our machine (e.g. open in your browser)? I assume this was already tested for other images but I don't fully remember

BrunoRosendo (changes requested):
good job, this is looking good! I got a few suggestions

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
AvilaAndre: I honestly don't remember this part, my guess is that I updated the branch and needed to modify it to include the gallery.

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
AvilaAndre: Changed the highlighted code as proposed, while keeping the companion objects.

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
AvilaAndre: It exists on the filesystem but the url does not work (I don't know if I am missing something because the port is 3000 and there is nothing there).

MRita443: @BrunoRosendo

BrunoRosendo: Please check #71 again, I'm pretty sure @DoStini already had the logic for this. There is a folder where Spring will save public files (see https://github.com/NIAEFEUP/website-niaefeup-backend/wiki/File-Upload#static-file-uploader)

I suspect the servePath is not being used correctly

AvilaAndre: Images are being stored locally in the `build/resources/main/static` folder, Spring Boot automatically serves the `src/main/resources/static` folder, however, due to the `endpointNotFound` present on the `ErrorController` (`src/main/kotlin/pt/up/fe/ni/website/backend/controller/ErrorController`) having a `RequestMapping("/**")`, it overrides the default behaviour and responds with an error 404 "invalid endpoint". Altough images are not being served, they also are not in the correct place for it, as they would not be served in the build directory.
To test this behaviour, remove the two `*` in the `RequestMapping` mentioned (leaving only `/`) and create an about.html with a `Hello World` string inside, you will see that it exists in `http://localhost:8080/about.html`.

src/main/kotlin/pt/up/fe/ni/website/backend/service/upload/CloudinaryFileUploader.kt
```diff
@@ -20,4 +20,11 @@ class CloudinaryFileUploader(private val basePath: String, private val cloudinar
 
         return result["url"]?.toString() ?: ""
     }
+
+    override fun deleteImage(fileName: String) {
+        cloudinary.uploader().destroy(
+            fileName,
```
MRita443, line 27: Shouldn't the file here be removed based on the the same path as in the upload, `"$basePath/$folder/$fileName"`?

src/main/resources/application.properties
```diff
@@ -32,9 +32,14 @@ upload.provider=static
...
 upload.cloudinary-url=GET_YOURS_AT_CLOUDINARY_DASHBOARD
 # Folder in which files will be stored
-upload.static-path=classpath:static
+upload.static-path=file:static
 # URL that will serve static content
-upload.static-serve=http://localhost:3000/static
+upload.static-serve=http://localhost:8080/static
+
+# Files are served in the following path
+spring.mvc.static-path-pattern=/static/**
```
MRita443, line 40: What does this do?

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -865,7 +865,7 @@ internal class EventControllerTest @Autowired constructor(
         fun `should fail if event does not exist`() {
             val unexistentID = 5
```
MRita443:
```suggestion
            val nonexistentID = 5
```

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -946,8 +947,8 @@ internal class EventControllerTest @Autowired constructor(
...
```
MRita443:
```suggestion
            val nonexistentID = 5
```

src/test/kotlin/pt/up/fe/ni/website/backend/utils/mockmvc/MockMvcMultipartBuilder.kt
```diff
@@ -49,6 +49,16 @@ class MockMvcMultipartBuilder(
...
+    fun asDeleteMethod(): MockMvcMultipartBuilder {
+        multipart.with(
+            {
+                it.method = "DELETE"
+                it
+            }
+        )
+        return this
+    }
+
```
MRita443:
```suggestion
   fun asDeleteMethod(): MockMvcMultipartBuilder {
        multipart.with {
            it.method = "DELETE"
            it
        }
        return this
    }
```

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -85,17 +83,9 @@ abstract class AbstractActivityService<T : Activity>(
...
-        var imageFolder = "activities"
-
-        if (activity is Event) {
-            imageFolder = EventService.IMAGE_FOLDER
-        } else if (activity is Project) {
-            imageFolder = ProjectService.IMAGE_FOLDER
-        }
-
         val fileName = fileUploader.buildFileName(image, activity.title)
         val imageName = fileUploader.uploadImage(imageFolder + "/gallery", fileName, image.bytes)
```
MRita443:
```suggestion
        val imageName = fileUploader.uploadImage("$imageFolder/gallery", fileName, image.bytes)
```

MRita443 (changes requested):
Some things:
- Tests are missing for `Project`
- Tests are missing the documentation functions

Rest is mostly questions and things my IDE nagged me about :)

src/main/kotlin/pt/up/fe/ni/website/backend/service/upload/CloudinaryFileUploader.kt
AvilaAndre: I'll add a comment above this line, but basically, it is called `fileName` due to the overriding the method in the supertype `FileUploader`, however, this is actually the `public_id` of the resource in Cloudinary, with will be the path. You can see this `public_id` in the method `uploadImage` above.

src/main/resources/application.properties
AvilaAndre: The first line I changed the port to be the same as where the server is running, because that is where the files will be located if we are using the local filesystem, the second line tells Spring where the static files to be served are.

MRita443:
I was just testing this out on Postman and noticed a couple more things
image

The image of the event itself has an incorrect port, perhaps we should fix that in this PR too.
Also, I think maybe using the slug would be better than the title for the filename

MRita443:
Also, regarding the documentation, I don't think it'll be possible for now because #144 is still open, my bad!

AvilaAndre:
> I was just testing this out on Postman and noticed a couple more things image
> 
> The image of the event itself has an incorrect port, perhaps we should fix that in this PR too. Also, I think maybe using the slug would be better than the title for the filename

Did that event already exist on your machine?
`3000` is the port that was hardcoded before. I already replaced that with 8080.

AvilaAndre:
> Also, regarding the documentation, I don't think it'll be possible for now because #144 is still open, my bad!

I already added some, the multipart is missing, maybe we just add that after #144 is finished.

Update: Already figured out why I can't use it.

MRita443:
> > I was just testing this out on Postman and noticed a couple more things image
> > The image of the event itself has an incorrect port, perhaps we should fix that in this PR too. Also, I think maybe using the slug would be better than the title for the filename
> 
> Did that event already exist on your machine? `3000` is the port that was hardcoded before. I already replaced that with 8080.

Yep you're right! Created another one and it's correct

MRita443 (approved):
Great job! This ended up being a bit more than expected but you handled it well, good eye catching all the slug mentions too :)

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -725,6 +812,195 @@ internal class EventControllerTest @Autowired constructor(
...
+                    jsonPath("$.image").value(testEvent.image)
+                )
+                /*
+                .andDocument(
+                    documentation,
+                    "Adds an image to the gallery of the selected event",
+                    "This endpoint adds an image to the gallery of an event, returning the event's updated data.",
+                    urlParameters = parameters
+                )
+                */
```
BrunoRosendo, line 873: is this not fixed yet? :'(

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
```diff
@@ -725,6 +812,195 @@ internal class EventControllerTest @Autowired constructor(
...
+                    jsonPath("$.image").value(testEvent.image)
+                )
+                /*
+                .andDocument(
+                    documentation,
+                    "Removes an image from the gallery of a selected event",
+                    "This endpoint removes an image to the gallery of an event, returning the event's updated data.",
+                    urlParameters = parameters
+                )
+                */
```
BrunoRosendo, line 967: same, others below

BrunoRosendo (approved):
Great job! :D I can't believe this is finally getting merged 😄

src/test/kotlin/pt/up/fe/ni/website/backend/controller/EventControllerTest.kt
MRita443: No, the `springdoc-openapi` doesn't support this yet :(

PR #204, opened 2024-01-04
Added time to DateInterval

### LINKED ISSUE #202: add event time to events
i'm trying to build the event details page and to do that i need not only the event date interval but also the event time. maybe it would be easier to change the date to a datetime.

jpnsantoss:
[ATTENTION. This is the *Pull Request Template*. Replace the two following lines with the indicated information and delete this one. Do not delete anything else]

Closes #202 

Added time to Date type from Date Interval, since it's needed for the events.

# Review checklist

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -17,7 +17,7 @@ server.error.whitelabel.enabled=false
 # Jackson
 spring.jackson.default-property-inclusion=non_null
 spring.jackson.deserialization.fail-on-null-creator-properties=true
-spring.jackson.date-format=dd-MM-yyyy
+spring.jackson.date-format=dd-MM-yyyy HH:mm:ss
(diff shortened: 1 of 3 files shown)


src/main/resources/application.properties
```diff
@@ -17,7 +17,7 @@ server.error.whitelabel.enabled=false
 # Jackson
 spring.jackson.default-property-inclusion=non_null
 spring.jackson.deserialization.fail-on-null-creator-properties=true
-spring.jackson.date-format=dd-MM-yyyy
+spring.jackson.date-format=dd-MM-yyyy HH:mm:ss
```
MRita443: Kind of nitty gritty here, but because the plan here is to use these for event time, maybe we could use hours and minutes only?

rubuy-74 (approved):
LGTM :rocket:

PR #215, opened 2025-02-12
feature: isActive account attribute

### LINKED ISSUE #212: account: add active attribute
We should add an `isActive` boolean to our accounts in order to distinguish active members from inactive ones. However, we should research beforehand if this can/should already be done with `Role`s.

peucastro:
Closes #212

This PR aims to add an `isActive` flag to the `Account` entity in order to distinguish active members from inactive ones.

# Review checklist


peucastro:
Set `isActive` as true and mandatory by default and correct tests.

src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
```diff
@@ -66,6 +67,9 @@ class Account(
     @JsonIgnore // TODO: Decide if we want to return roles (or IDs) by default
     val roles: MutableList<@Valid Role> = mutableListOf(),
 
+    @field:NotNull
+    var isActive: Boolean = true,
+
```
MRita443: From what we discussed and the fact that this attribute is mandatory on the creation dto, it's a bit redundant to have a default, no harm in leaving it though

src/test/kotlin/pt/up/fe/ni/website/backend/utils/documentation/payloadschemas/model/PayloadAccount.kt
```diff
@@ -52,6 +52,12 @@ class PayloadAccount(includePassword: Boolean = true) : ModelDocumentation(
             optional = true,
             isInRequest = false
         ),
+        DocumentedJSONField(
+            "isActive",
+            "Flag that indicated whether the account is active or inactive",
```
MRita443:
```suggestion
            "Flag that indicates whether the account is active or inactive",
```

MRita443 (approved):
Looks good to me! Some non-essential comments but we can also close as is!

MRita443 (changes requested):
I just noticed, we should add `isActive` value checks to the rest of the tests in 
`AccountControllerTest` to be thorough

MRita443 (approved):
Looks great!

PR #221, opened 2025-03-02
Fix/delete unused files

### LINKED ISSUE #177: files: delete files when not being used
Currently, accounts, projects and events are storing their images even if the entities are deleted or their images are changed. We should delete them in these cases in order to save space.

PedroOliveira42:
Closes #177 

Added and tested code to delete unused files in the backend
# Review checklist
-   [X] Behavior is as expected
-   [X] Clean, well structured code
-   [X] Properly documents API changes in the corresponding controller test
-   [X] Contains enough appropriate tests (due to fileuploader nature only tested manually)
-   [X] Changes are reflected in the Wiki if necessary (No change was assumed necessary)


src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -71,8 +74,10 @@ class AccountService(
     }
 
     fun deleteAccountById(id: Long) {
-        if (!repository.existsById(id)) {
-            throw NoSuchElementException(ErrorMessages.accountNotFound(id))
+        val account = getAccountById(id)
+
+        account.photo?.takeIf { it.isNotBlank() }?.let { photo ->
+            fileUploader.deleteImage(photo)
```
MRita443: So this looks good right, it first checks if the account photo is not null, and then if it's not an empty string. Only then it deletes the photo. But I don't think it's needed because we have the `    @field:NullOrNotBlank` annotation on the `photo` field, so it will never be blank, only null

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/AbstractActivityService.kt
```diff
@@ -106,4 +109,13 @@ abstract class AbstractActivityService<T : Activity>(
...
     }
+
+    protected fun removeAllActivityImages(activityImage: String, activityGallery: MutableList<String>) {
+        fileUploader.deleteImage(activityImage)
+        activityGallery.forEach { imageName ->
+            if (imageName.isNotBlank()) {
+                fileUploader.deleteImage(imageName)
+            }
+        }
+    }
```
MRita443:
```suggestion
    protected fun removeAllActivityImages(activityImage: String, activityGallery: MutableList<String>) {
        fileUploader.deleteImage(activityImage)
        activityGallery.forEach { imagePath ->
            if (imagePath.isNotBlank()) {
                fileUploader.deleteImage(imagePath)
            }
        }
    }
```

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/EventService.kt
```diff
@@ -38,10 +38,11 @@ class EventService(
     }
 
     fun deleteEventById(eventId: Long) {
-        if (!repository.existsById(eventId)) {
-            throw NoSuchElementException(ErrorMessages.eventNotFound(eventId))
-        }
-
+        val event = getEventById(eventId)
+            ?: throw NoSuchElementException(ErrorMessages.eventNotFound(eventId))
```
MRita443:
```suggestion
```

The Elvis operator is not needed because `getEventById` does that already :)

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/ProjectService.kt
```diff
@@ -51,11 +51,12 @@ class ProjectService(
     }
 
     fun deleteProjectById(id: Long) {
-        if (!repository.existsById(id)) {
-            throw NoSuchElementException(ErrorMessages.projectNotFound(id))
-        }
-
+        val project = getProjectById(id)
+            ?: throw NoSuchElementException(ErrorMessages.projectNotFound(id))
```
MRita443:
```suggestion
```
Same here

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -49,8 +49,11 @@ class AccountService(
         }
 
         dto.photoFile?.let {
+            val previousPhoto = account.photo
             val fileName = fileUploader.buildFileName(it, dto.email)
             account.photo = fileUploader.uploadImage("profile", fileName, it.bytes)
+            previousPhoto?.let { fileUploader.deleteImage(it) }
+            // This seems like it should be inside a try catch, but no db changes are, ask in next meet
```
MRita443: I don't think this is needed because that kind of treatment is already done either in the `deleteImage` function or in the functions it calls

MRita443 (changes requested):
Great job! Really minor things

rubuy-74 (approved):
Just tested it and it works perfectly. Nice work :rocket:

MRita443 (approved):

PR #228, opened 2025-06-30
Mandatory event end date

Isaac3105:
Closes #273

This PR is only complementary to PR #313 from website-niafeup-frontend, to fix the unnecessary use of endDate as optional in tests and in general use for forms and such.

# Review checklist
-   [x] Changes are reflected in the Wiki if necessary


MRita443 (approved):
Looks great to me! Just for future explanation, we made `endDate` mandatory because as Dates now include hours, every event will have an end date, even if it's just a couple hours after the start date

PR #234, opened 2025-07-22
Set content security policy

### LINKED ISSUE #233: XSS protection
As described here: https://www.baeldung.com/spring-prevent-xss

MRita443:
Closes #233 

Sets content security policy as indicated here: https://www.baeldung.com/spring-prevent-xss.
Did not use the X-XSS-Protection header as that is now deprecated.

# Review checklist


src/main/kotlin/pt/up/fe/ni/website/backend/config/security/AuthConfig.kt
```diff
@@ -35,6 +35,15 @@ class AuthConfig(
         http {
             csrf { disable() }
             cors { }
+            headers {
+                contentSecurityPolicy {
+                    policyDirectives =
+                        "default-src 'self';" +
+                        " script-src 'self';" +
+                        " style-src 'self' 'unsafe-inline';" +
+                        " img-src 'self' data: https:"
```
peucastro: We should consider adding additional CSP directives for stronger XSS protection, such as [`form-action`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/form-action), [`base-uri`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/base-uri) and so on. Notice that most of them are already set by [`default-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src), but it doesn’t cover all scenarios.

src/main/kotlin/pt/up/fe/ni/website/backend/config/security/AuthConfig.kt
```diff
@@ -35,6 +35,15 @@ class AuthConfig(
         http {
             csrf { disable() }
             cors { }
+            headers {
+                contentSecurityPolicy {
+                    policyDirectives =
+                        "default-src 'self';" +
+                        " script-src 'self';" +
+                        " style-src 'self' 'unsafe-inline';" +
```
peucastro: As stated in the [mdn reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#inline_javascript):

> Developers should avoid 'unsafe-inline', because it defeats much of the purpose of having a CSP. Inline JavaScript is one of the most common XSS vectors, and one of the most basic goals of a CSP is to prevent its uncontrolled use.

Unless there’s a specific need for **`unsafe-inline`** here, it’s safer to remove it.

peucastro (approved):
✅

PR #248, opened 2025-10-11
179 activities change team relationship to manytomany

### LINKED ISSUE #179: activities: change team relationship to ManyToMany
Currently, the teamMembers and hallOfFame fields of the `Activity` entity hold a oneToMany relationship. This is incorrect because an account can be part of multiple teams.

We should change the relationship to ManyToMany

Vidal322:
Closes #179 

# Review checklist


src/main/kotlin/pt/up/fe/ni/website/backend/model/Account.kt
```diff
@@ -40,6 +40,11 @@ class Account(
     @field:Size(min = Constants.Password.minSize, max = Constants.Password.maxSize)
     var password: String,
 
+    @ManyToMany(mappedBy = "teamMembers", fetch = FetchType.EAGER)
+    // @JsonIdentityReference(alwaysAsId = true)
```
peucastro:
```suggestion
```

src/main/kotlin/pt/up/fe/ni/website/backend/model/Activity.kt
```diff
@@ -29,8 +32,13 @@ abstract class Activity(
...
 
-    @JoinColumn
-    @OneToMany(fetch = FetchType.EAGER)
+    @ManyToMany(fetch = FetchType.EAGER)
+    @JoinTable(
+        name = "activity_team_members",
+        joinColumns = [JoinColumn(name = "activity_id")],
+        inverseJoinColumns = [JoinColumn(name = "account_id")]
+    )
+    @OrderColumn
```
peucastro, line 41: Ideally, both `teamMembers` in the `Activiy` model and the `hallOfFame` in the `Project` model should be of type `MutableSet`. This way we won't need the `@OrderColumn` annotation. This can be a little tricky to do, but it will increase the performance, since it will eliminate these unnecessary index management in the database.

peucastro (changes requested):

peucastro (approved):
Thank you for this! 🫡 ✅ ✈️

PR #252, opened 2025-10-25
Icon path to icon picture

### LINKED ISSUE #251: Use MultipartFile for CustomWebsiteDto.iconPath
Replace the `iconPath` field in `CustomWebsiteDto` with a `MultipartFile` to support direct file uploads instead of passing a URL path.

Also update the controller and service layers to handle file upload and deletion accordingly.

Isaac3105:
Closes #251

Migration from receiving and handling a URL to a icon on the customWebsite entity to using a multipartFile.

# Review checklist


Isaac3105:
The current version has a quick fix for the tests. 
In the file customWebsiteDto, I added:
```
@field:NullOrNotBlank
var iconPath: String? = null,
```
These lines would only be used to pass the tests however, without them about 5 or 6 tests would fail. 
There is possibly a better option for this.

peucastro (changes requested):
Overall, this look pretty good to me! However, the `Project` entity also uses `CustomWebsite` objects in its links field, but `ProjectController` and `ProjectService` haven't been updated to handle icon uploads for those links. To complete this feature, you'll need to add a similar `linkIcons` parameter to the project create/update endpoints and implement the corresponding file handling logic in `ProjectService`, mirroring what was done for accounts. Additionally, the seeders still reference URL-based `iconPaths` that should be updated to null or use actual file uploads.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -38,9 +38,16 @@ class ProjectController(private val service: ProjectService) {
...
         @RequestParam
         @ValidImage
-        thumbnail: MultipartFile
+        thumbnail: MultipartFile,
+        @RequestParam
+        @ValidImage
+        linkIcons: MutableList<MultipartFile?>?
     ): Project {
         project.imageFile = image
+        project.thumbnailFile = thumbnail // não sei se é necessário
```
peucastro: Yes, it is! 😄

src/main/kotlin/pt/up/fe/ni/website/backend/controller/ProjectController.kt
```diff
@@ -38,9 +38,16 @@ class ProjectController(private val service: ProjectService) {
         image: MultipartFile,
         @RequestParam
         @ValidImage
-        thumbnail: MultipartFile
+        thumbnail: MultipartFile,
+        @RequestParam
+        @ValidImage
+        linkIcons: MutableList<MultipartFile?>?
```
peucastro:
```kt
@JsonIgnore
var linkIconFiles: MutableList<MultipartFile?>? = null
```

This should be added to `ProjectDto` to properly handle the list of icon files, similar to how `AccountDto` has `websiteIconFiles`.

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/ProjectService.kt
```diff
@@ -72,6 +85,19 @@ class ProjectService(
...
+            val dtoWebsite = dto.links?.getOrNull(index)
+            val iconFile = dtoWebsite?.iconFile
+            if (iconFile != null && !iconFile.isEmpty) {
+                website.iconPath?.let { previousIcon ->
+                    fileUploader.deleteImage(previousIcon)
+                }
+                val fileName = fileUploader.buildFileName(iconFile, dto.slug)
+                website.iconPath = fileUploader.uploadImage("customWebsiteIcon", fileName, iconFile.bytes)
+            }
+        }
```
peucastro: This logic relies on `dtoWebsite.iconFile` being set in the controller. A better approach would be following the same pattern as `AccountService` and iterate through `dto.linkIconFiles` by index, something like this:

```kt
for ((index, website) in project.links.withIndex()) {
    val iconFile = dto.linkIconFiles?.getOrNull(index)
    if (iconFile != null && !iconFile.isEmpty) {
        val fileName = fileUploader.buildFileName(iconFile, dto.slug)
        website.iconPath = fileUploader.uploadImage("customWebsiteIcon", fileName, iconFile.bytes)
    } else {
        website.iconPath = null
    }
}
```

src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/account/CreateAccountDto.kt
peucastro, line 1: Before we get this merged, please recheck if all the changes made by the annotations match with the requirements from the frontend side.

src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/account/UpdateAccountDto.kt
peucastro, line 1: Before we get this merged, please recheck if all the changes made by the annotations match with the requirements from the frontend side.

peucastro (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/ProjectService.kt
Isaac3105: I left it like this:
```kotlin
for ((index, website) in newProject.links.withIndex()) {
    val iconFile = dto.linkIconFiles?.getOrNull(index)
    if (iconFile != null && !iconFile.isEmpty) {
        website.iconPath?.let { previousIcon ->
            fileUploader.deleteImage(previousIcon)
        }
        val fileName = fileUploader.buildFileName(iconFile, dto.slug)
        website.iconPath = fileUploader.uploadImage("customWebsiteIcon", fileName, iconFile.bytes)
    }
}
``` 

so that if the image is null (user doesn't actually update the image) we use the previous image.

Isaac3105: sooo, both of us were wrong... after testing it more carefully, i saw that if an account had multiple websites and I only updated one icon, the first website would have that new icon and all the other ones would loose their icons.... I think I really fixed that issue now but I had to goo trough a long way to make it happen, I'm not sure if it is the best way possible tho.

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
```diff
@@ -29,26 +29,35 @@ class AccountController(private val service: AccountService) {
     @GetMapping("/{id}")
     fun getAccountById(@PathVariable id: Long) = service.getAccountById(id)
 
+    private fun extractWebsiteIcons(request: MultipartHttpServletRequest): Map<String, MultipartFile> {
+        return request.fileMap
+            .filter { it.key.startsWith("websiteIcons[") }
+    }
```
peucastro: Isn't it possible to use a `@RequestParam` attribute in the controller endpoints instead of creating an auxiliary function?

src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/account/UpdateAccountDto.kt
```diff
@@ -1,21 +1,42 @@
...
 import java.util.Date
 import org.springframework.web.multipart.MultipartFile
 import pt.up.fe.ni.website.backend.dto.entity.CustomWebsiteDto
 import pt.up.fe.ni.website.backend.dto.entity.EntityDto
 import pt.up.fe.ni.website.backend.model.Account
+import pt.up.fe.ni.website.backend.utils.validation.NullOrNotBlank
 
 class UpdateAccountDto(
-    val email: String,
+    @field:NotBlank
```
peucastro: The name should not be required in every `PUT` request

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/ProjectService.kt
```diff
@@ -44,6 +44,19 @@ class ProjectService(
...
+            if (customWebs == null) continue
+            val dtoWebsite = customWebs.getOrNull(index) ?: continue
+            val iconFile = dtoWebsite.iconFile
+            if (iconFile != null && !iconFile.isEmpty) {
+                val fileName = fileUploader.buildFileName(iconFile, dto.slug)
+                website.iconPath = fileUploader.uploadImage("customWebsiteIcon", fileName, iconFile.bytes)
+            } else {
+                website.iconPath = null
+            }
+        }
```
peucastro: This logic relies on the `project.links` and `dto.links` lists having the exact same order and size, which is fragile. Additionally, the `else` block will accidentally overwrite existing icons with `null` during updates if a new file isn't uploaded.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -51,19 +59,46 @@ class AccountService(
...
+                val fileName = fileUploader.buildFileName(iconFile, dto.email)
+                website.iconPath = fileUploader.uploadImage("customWebsiteIcon", fileName, iconFile.bytes)
+            } else {
+                website.iconPath = oldWebsites.getOrNull(index)?.iconPath
+            }
+        }
+
+        oldWebsites.drop(newAccount.websites.size).forEach { oldWebsite ->
+            oldWebsite.iconPath?.let { fileUploader.deleteImage(it) }
+        }
```
peucastro: The index-based matching between `oldWebsites` and `newAccount.websites` is fragile and can cause data corruption if links are reordered or deleted from the middle of the list. Instead of using drop, you should identify removed links by comparing IDs to ensure the correct icons are deleted. I recommend passing a unique identifier in the `websiteIcons` map keys to safely match files to entities.

peucastro (changes requested):

src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/account/UpdateAccountDto.kt
```diff
@@ -1,21 +1,42 @@
...
+    @field:Valid
+    val websites: List<CustomWebsiteDto>?
+) : EntityDto<Account>(Account::class) {
+
+    @JsonIgnore
+    var photoFile: MultipartFile? = null
+
+    @JsonIgnore
+    var websiteIconFiles: Map<String, MultipartFile>? = null
+}
```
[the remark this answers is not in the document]
Isaac3105: noted. this is indeed wrong. I will fix it in my next commits

src/main/kotlin/pt/up/fe/ni/website/backend/controller/AccountController.kt
Isaac3105: its not :(. because the user can have multiple websites, if he has 3 for example and only update the third website icon, it is necessary to have a index to know which website he intends to put the image. since request param cant parse into indexed maps (website number and image), it really is necessary to do this workaround I think.

Isaac3105: at least all of the ways I tried it failed.

src/main/kotlin/pt/up/fe/ni/website/backend/dto/entity/account/UpdateAccountDto.kt
Isaac3105: right. I'll fix it in my next commit.

src/main/kotlin/pt/up/fe/ni/website/backend/service/activity/ProjectService.kt
Isaac3105: I think this is already taken care of.

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
Isaac3105: what kind of unique identifier tho...? I'll try to maybe pass the url....

Isaac3105: ok did that, it went well id... i mean, if the user changes the url, it forgets the previous icon tho. i can't really see how it could be done with another type of unique identifier...

Isaac3105: turns out, its not possible to use a id because the ids are automatically generated values in the backend, and we can't create a new id in the frontend and reuse them in the backend.

peucastro (commented):
As we talked offline, to simplify frontend interactions for managing individual custom websites, i recommend doing this (all of these tips are susceptible to changes and discussion :) )

Implement `addCustomWebsite` and `removeCustomWebsite` endpoints (and potentially `updateCustomWebsite`) within `AccountController` and `ProjectController`. This will allow the frontend to easily add or remove a single custom website without resubmitting the entire list.

Consider retaining the existing multipart request mechanism in the create and update operations for `Account` and `Project` if it works. This approach remains efficient for initial creation.

Also please ensure to review the git diff and pay attention to the relevance and necessity of the included changes.

Isaac3105:
Did what you said. tested it with the user edit page and it worked out perfectly. however, I will say there is a frontend issue. please go there to tell me what you think... @peucastro

Isaac3105:
removed the outdated websites arrays and assertions from the main create/update payloads in AccountControllerTest since that data is no longer handled there

The ProjectDto, ProjectController, and ProjectService still had old code trying to map linkIconFiles from multipart requests so I removed it all

In tests: Removed the StandardMultipartHttpServletRequest injection from the Project create/update endpoint signatures. (It was only there for the legacy link parsing, and MockMvc doesn't support it, which was causing the 500 Internal Server Error crashes during testing). Removed the legacy links parameter injections, size assertions, and nested LinksValidation checks from ProjectControllerTest

after all that I tested it with the user info page branch frontend and all is still working great.

PR #255, opened 2025-12-08
thumbinail Backend fixed

DaltonDjoesman:


peucastro (approved):
Nice work in the `ProjectConrtoller`! However, I have noticed that the image handling in `updateProjectById` has a logic error: when a new thumbnail is uploaded, the service deletes `project.image` (the hero image) instead of `project.thumbnail`. We need to manage the lifecycle of both images independently, so that replacing one doesn't delete the other

peucastro (approved):
Thank you! ✅

PR #260, opened 2025-12-23
feat: admin created accounts receive email with jwt token to set password

### LINKED ISSUE #83: accounts: Create a new account
We should consider how accounts will be created when the website is launched. I can think of two ways to do this:
- Generate some kind of token (jwt or a simple validation token stored in the DB) and send it in a link to the new member's email. In this implementation, the account information and email would be retrieved from the application. Since we don't have applications yet, we could have a simple endpoint that creates an invalidated account with no password (with permission to do so). A second endpoint is needed to associate the password to the account and validate the token.
- Have an endpoint to create a new account (with special permission) with a random password that is sent by email. Later, that person could change their password. This is similar to what we currently have

These are just 2 options, we should further discuss this matter and the options we have

tomaspalma:
Closes #83 

The email could be more pretty, but I don't think we have anything from the image department and it is also not that important maybe

image

## Made changes

- Creates a JWT token using the app's private and public RSA key pair
- Sends an email to the user
- Added a `docker-compose.yaml` with mailpit to better test emails
- Creating an account does not need a password anymore
- Created email template for signed urls

# Review checklist


docker-compose.yaml
```diff
@@ -0,0 +1,7 @@
+services:
+  mailpit:
+    image: axllent/mailpit
+    container_name: tts-mailpit
```
peucastro: oopsie 😅 

```suggestion
    container_name: website-niaefeup-mailpit
```

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -24,13 +31,29 @@ class AccountService(
...
+                    SignedURLEmailDto(
+                        "A tua conta foi criada",
+                        "Criámos a tua conta para acederes ao site do NIAEFEUP!",
+                        "Clica aqui para definires a tua password",
+                        signedUrl,
+                        authProperties.jwtAccessExpirationMinutes
+                    )
+                )
+                .to(dto.email)
+                .from("ni@aefeup.pt")
```
peucastro: Here we can inject a `@Value("\${email.from}") private val emailFrom: String` into the service parameters to reuse the value defined in `application.properties`

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
```diff
@@ -24,13 +31,29 @@ class AccountService(
...
 
         val account = dto.create()
-        account.password = encoder.encode(dto.password)
 
         dto.photoFile?.let {
             val fileName = fileUploader.buildFileName(it, dto.email)
             account.photo = fileUploader.uploadImage("profile", fileName, it.bytes)
         }
 
+        val signedUrl = signedURLService.generateSignedUrl("/set-password", CreateAccountJWTClaimDto(dto.email))
```
peucastro: This needs to be updated as soon as we have a functional "Recover password" in the frontend

src/main/resources/templates/email/layout.html.mustache
```diff
@@ -5,7 +5,7 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <title>{{{subject}}}</title>
+
```
peucastro, line 8:
```suggestion
    <title>{{{subject}}}</title>
```

peucastro (changes requested):

src/main/resources/templates/email/layout.html.mustache
tomaspalma: Hi @peucastro, what does this change exactly? It looks the same to me

src/main/kotlin/pt/up/fe/ni/website/backend/service/AccountService.kt
tomaspalma: Apparently, the `from` is already defined in the `BaseEmailBuilder` and it already comes from `application.properties`, so I can just not have the `.from("ni@aefeup.pt") here.

src/main/resources/templates/email/layout.html.mustache
peucastro: I was just wondering why this line `<title>{{{subject}}}</title>` was removed

PR #263, opened 2026-01-29
chore: created dotenv environment for production

### LINKED ISSUE #262: config: Manage .env environment in application.properties
Instead of directly passing the value to the `application.properties` file, it would be better to pass them as environment variables (e.g. `spring.mail.host=${SPRING_MAIL_HOST}`).

Using environment variables is the standard way we manage secrets and configurations across our organization, and it's how Niployments expects to.

Investigate what is the best way to do this, the upsides and downsides of using something like [dotenv-java](https://github.com/cdimascio/dotenv-java).

pedroafmonteiro:
Closes #262 

- Changed `application.properties` to support variables for the necessary configurations.
- For local development, users will need to copy `.env.example` as `.env.local` to begin developing the project. Users can modify `.env.local` as they would like (e.g. inserting mail credentials to test the email system).
- For production, secrets will be injected by Kubernetes Secrets directly via system environment variables, so no need for dotfiles in the project root in a production environment.

# Review checklist


tomaspalma:
In projects such as nijobs ( :dove:  ) and tts, in order prevent someone changing the value of a variable because locally it's best for them and then comitting that into github which will affect other people, it's a common practice to have the env file in the `.gitignore` and then having a `.env.example` file with default values. 

For example, If someone wants to use mailpit to test their emails, and then someone in their pr changes the `.env.local` variables to test with the real world scenario (to test something, for instance), and then commits that and gets merged to develop, then other people that normally used mailpit would have to then change the values of those variables. Also, it would require the person who changed the env variables in order to test emails with something else to make sure they dont commit that file. If the actual env file was in the `.gitignore`, they would not need to worry about this.

Developers will need to create their own `.env.local` file with the contents of `.env.example` only during the first time they clone the repository.

Additionally, if, for some reason, someone uses a real key or something to test something locally and puts that into `.env.local` and then accidentally commits that into the repository, there is a real danger the secret is exposed.

Let me know what you think or if I badly explained something

pedroafmonteiro:
hey @tomaspalma, i've just done your suggestions. tell me what you think :)

tomaspalma (approved):

peucastro (approved):
Insane changes 🔥 🔥

